PR #9, opened 2022-10-18
chore: init project

homura:
## Description

Init Kuai project, with a `@kuaiproject/cli` package for demo

```sh
git clone https://github.com/homura/kuai.git
cd kuai
yarn
yarn build
# workspace symlink
yarn install --force
yarn kuai node --port 8114
```


Development.md
```diff
@@ -0,0 +1,15 @@
+# Development
+
+## Quick Start
+
+```sh
+git clone https://github.com/ckb-js/kuai.git
+cd kaui
```
Keith-CY, line 7:
```suggestion
cd kuai
```

packages/cli/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "@kauiproject/cli",
```
Keith-CY, line 2:
```suggestion
  "name": "@kuaiproject/cli",
```

Keith-CY:
I'm not sure if it's a good time to start coding because we don't have production and architecture design.

Besides, I think `running a node` via `task` would be better than embedding it in the `cli` module

.eslintignore
```diff
@@ -0,0 +1 @@
+*.js
```
Keith-CY, line 1: newline at the end of file is required

.gitignore
```diff
@@ -0,0 +1,9 @@
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
+lerna-debug.log*
+
+node_modules/
+**/lib
+
+.idea
```
Keith-CY, line 9: newline at the end of file is required

.prettierrc
```diff
@@ -0,0 +1,6 @@
+{
+  "printWidth": 120,
+  "trailingComma": "all",
+  "singleQuote": true,
+  "arrowParens": "always"
+}
```
Keith-CY, line 6: newline at the end of file is required

homura:
> I'm not sure if it's a good time to start coding because we don't have production and architecture design

this PR just init the project structure, and, I think that we probably already have some more deterministic things that we can start coding, such as 
- fetch a ckb container from DockerHub
- a `ckb.toml` for launching local network

> Besides, I think `running a node` via `task` would be better than embedding it in the `cli` module

We do need to design Task first, the cli package for this PR is for demo only, not directly for production

[Task](https://github.com/NomicFoundation/hardhat/blob/1282f8c6ee6ea8c84a5e596903f40a765ea4d5f8/packages/hardhat-core/src/internal/core/tasks/task-definitions.ts) or [Command](https://github.com/tj/commander.js/blob/master/lib/command.js) are actually quite similar, commander's `Command` should be more popular than Hardhat task (according to the number of npm downloads), so I'll use it as an example

Keith-CY:
There're two conventions we'd better follow
1. pin dependencies for clarity and certainty;
2. sign every commit for security.

Keith-CY (approved):

Development.md
Keith-CY: The typo `kaui` is not fixed, why is the conversation marked as resolved?

homura: I made a mistake and realized that two "kuai" had been typed as "kaui", I applied this suggestion

yanguoyu:
Need to add README.md for users.

yanguoyu (approved):

PainterPuppets (approved):

Issue #13, opened 2022-11-01
Design of task/plugin system


PainterPuppets:
# The design of kuai task system 

## 1. Why we need a task system
As a build system, kuai needs to be able to perform tasks such as compiling, testing, starting nodes, etc. To better manage and expand these tasks, kuai needs a task system. In order to manage and expand these tasks, kuai needs a task system.

## 2. Design target (What)

To achieve the above goals, the task system needs to have the following features
 - dynamically load tasks to facilitate expansion/new custom tasks
 - The task runtime needs to be like an onion model to implement pre/post-processing operations when multiple task handlers are used
 - has a `KuaiContext`, which may have config/anything, usually the context is a singleton (providing the ability to create multiple contexts manually)
 - task can be executed directly from the cli via `npx kuai TASK_NAME`
 - task can be executed manually in the code to facilitate doing e2e tests
 - support subtask

## 3. How it works
Clarifying how many modules there are and what the system architecture looks like first helps us understand how the system works

- `KuaiContext`: Load configurations, plugins, tasks, etc. and initialize other modules.
- `TaskSystem`
  - `RuntimeEnvironment`: Execution/management of task status, and task lifecycle.
  - `Task`: Specific logic
  - `OverrideTask`: extends from `Task`, for override task
- `Kuai CLI`: Parsing command line arguments, and providing good cli interaction and output.

```mermaid
  C4Component
  title Architecture of kuai project context
    
  Container(ctx, "Kuai Context", "", "Load configurations, plugins, tasks, etc. and initialize other modules.")
  Container(cli, "CLI", "", "Parsing command line arguments, and providing good cli interaction and output.")

  Container_Boundary(st, "Task system") {
      Component(kre, "Kuai Runtime Environment", "", "Execution/management of task status, and task lifecycle.")
      Component(task, "Task", "", "Specific logic")
      Component(otask, "OverrideTask", "", "Task override")

      Rel(kre, task, "Manage & Run")
      Rel(otask, task, "Override")
  }

  Rel(ctx, kre, "init")
  Rel(cli, ctx, "call")
```

### 3.1. KuaiContext, the entry of the kuai

`KuaiContext` is the entry point for the whole system, responsible for loading config, task, etc... And to prevent multiple `KuaiContext` at runtime, it should be a singleton pattern, and a new instance can be created by `KuaiContext.createKuaiContext()`.

```mermaid
  classDiagram
  class KuaiContext {
    + RuntimeEnvironment env
[...]

PR #14, opened 2022-11-02
docs: kuai task system design doc

PainterPuppets:
This pr contains a task system design document, create it just for easy checking of changes

preview: https://github.com/PainterPuppets/kuai/blob/kuai_task_system_design_doc/task-system-design.md
resolve: https://github.com/ckb-js/kuai/issues/13

### DIFF
--- a/task-system-design.md
+++ b/task-system-design.md
@@ -1,7 +1,9 @@
-## kuai task system design
+## The design of kuai task system 
 
-### design target
-The `kuai` project needs a task system to support inject dapp template / run node / deploy contract etc...
+### 1. Why we need a task system
+As a build system, kuai needs to be able to perform tasks such as compiling, testing, starting nodes, etc. To better manage and expand these tasks, kuai needs a task system. In order to manage and expand these tasks, kuai needs a task system.
+
+### 2. Design target (What)
 
 To achieve the above goals, the task system needs to have the following features
  - dynamically load tasks to facilitate expansion/new custom tasks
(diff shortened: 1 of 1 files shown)


task-system-design.md
```diff
@@ -0,0 +1,268 @@
...
+  }
+```
+
+**Possible usage**
+```ts
+// entry.ts
+import { KuaiContext, RuntimeEnvironment } from 'kuai/core'
+
+const ctx = HardhatContext.createHardhatContext();
+const config = loadConfigAndTasks();
```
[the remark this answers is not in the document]
IronLu233: But the `RuntimeEnvironment` class constructor seems like `constructor(config, tasks, extenders)`. So I think `loadCofigAndTasks ` may be renamed to `loadConfig`

task-system-design.md
```diff
@@ -0,0 +1,268 @@
...
+
+task(TASK_NAME)
+  .addParam("paramA", "a custom param", 'default value')
+  .action(async (args, env, runSuper) => {
+  // do something…
+  })
+
+// kuai.config.ts
+
+import 'custom-plugin.ts'
```
IronLu233, line 157: `import 'custom-task'` ?
I can not find the context of `plugin`

task-system-design.md
```diff
@@ -0,0 +1,268 @@
...
+
+```ts
+// custom-override.ts
+
+import { task } from 'kuai/core'
+
+task(TASK_NAME)
+  .action(async (args, env, runSuper) => {
+    // do something before parent task…
+    await runSuper()
```
IronLu233, line 170: is the `runSuper` nullable?, If it is, tell the user only available when it is overrided.

IronLu233 (commented):
I think these things can be more clearly:
1. the sequence diagram when user use this task system via cli
2. How the RuntimeEnvironment find and run the tasks(it seems need a task map from task name to task and override tasks)

PR #25, opened 2022-11-15
fix: fix configuration

Keith-CY:
1. rename "kuaiproject" to "kuai"
2. use `npm` by default cuz `yarn` is not configured in `lerna.json`
3. fix inconsistent versions among packages and project
4. set node.js@18 required cuz 18 is LTS and 16 will be deprecated in a year
5. rename the "CI" action to "Test" which is more explicit


Keith-CY:
Since we're using `lerna` to manage packages, the dependencies and versions should be delegated to `lerna` by `lerna add` and `lerna version`

PainterPuppets (approved):

yanguoyu (approved):

PR #32, opened 2022-11-20
feat: integrate cli

PainterPuppets:
integrate kuai/cli & kuai/core

- [x] use kuai task system by cli
- [x] implement `help` command 

image

task system design: https://github.com/ckb-js/kuai/issues/13

### DIFF
--- a/packages/cli/src/main.ts
+++ b/packages/cli/src/main.ts
@@ -1,8 +1,94 @@
 #!/usr/bin/env node
-import { Command } from 'commander'
+import { Command, createOption, Option } from 'commander'
 import { node } from './commands/node'
+import { TaskParam, TaskArguments, initialKuai, paramTypes, KuaiArguments } from '@kuai/core'
 
-const program = new Command()
+const KUAI_GLOBAL_PARAMS: Array<TaskParam> = [
+  {
+    name: 'config',
+    description: 'A Kuai config file.',
+    type: paramTypes.path,
+    isFlag: false,
+    isOptional: true,
+  },
+]
 
-program.addCommand(node)
-program.parse()
+function parseTaskParams(param: TaskParam<TaskArguments>): Option {
+  const type = param.type.name === 'boolean' ? '' : ` <${param.type.name}>`
+
+  return createOption(`--${param.name}${type}`, param.description)
+}
+
+const main = async () => {
+  const program = new Command()
+
+  KUAI_GLOBAL_PARAMS.map((param) => parseTaskParams(param)).forEach((option) => program.addOption(option))
+
+  program.addCommand(node)
+
+  const args: KuaiArguments = (() => {
+    const result: KuaiArguments = {}
+
+    const configPathFlagIndex = process.argv.slice(0, -1).lastIndexOf('--config')
+
+    if (configPathFlagIndex !== -1) {
+      Object.assign(result, { configPath: process.argv[configPathFlagIndex + 1] })
+    }
+
+    return result
+  })()
+
+  const ctx = await initialKuai(args)
+  const env = ctx.getRuntimeEnvironment()
+  Object.values(env.tasks)
+    .filter((task) => !task.isSubtask)
+    .forEach((task) => {
+      const cmd = new Command(task.name)
+
+      if (task.description) {
+        cmd.description(task.description)
+      }
+
+      Object.values(task.params)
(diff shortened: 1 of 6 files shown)


packages/cli/__tests__/__fixtures__/kuai-config-case/kuai.config2.ts
```diff
@@ -0,0 +1,13 @@
...
+  .addParam('paramA', 'a custom param', 'default value')
+  .setAction(() => {
+    console.log('demo-task2');
+  });
+
+const config: KuaiConfig = {
+  kuai: '0.0.1',
+};
+
+export default config;
```
yanguoyu, line 13: `demo-task1` is similar to `demo-task2`, I think for the test case it only needs to leave one.

PainterPuppets: These two files are designed to test that "different commands are loaded when different configuration files are specified"

Maybe it would be better to make the two files look a more different?

yanguoyu: Add some commands in a configuration, and test if it will load. I feel it's enough.

felicityin (approved):

packages/cli/src/main.ts
```diff
@@ -1,8 +1,68 @@
...
+  },
+];
+
+function parseTaskParams(param: TaskParam<TaskArguments>): Option {
+  return createOption(`--${param.name} <${param.type.name}>`, param.description);
+}
 
 const program = new Command();
 
+KUAI_GLOBAL_PARAMS.map((param) => parseTaskParams(param)).map((option) => program.addOption(option));
```
Keith-CY, line 23:
```suggestion
KUAI_GLOBAL_PARAMS.map((param) => parseTaskParams(param)).forEach((option) => program.addOption(option));
```

packages/cli/src/main.ts
```diff
@@ -1,8 +1,68 @@
...
+  let path;
+  for (let i = 0; i < process.argv.length; i++) {
+    if (process.argv[i] === '--config') {
+      if (i == process.argv.length - 1) {
+        throw new KuaiError(ERRORS.ARGUMENTS.MISSING_CONFIG_FILE);
+      }
+      path = process.argv[i + 1];
+      break;
+    }
+  }
```
Keith-CY, line 37: `process.argv.slice(0, -1).lastIndexOf('--config')` should work to check if the `config file` is specified.

packages/cli/src/main.ts
```diff
@@ -1,8 +1,100 @@
...
+
+  program.addCommand(node);
+
+  const args: KuaiArguments = (() => {
+    const result: KuaiArguments = {};
+
+    const configPathFlagIndex = process.argv.slice(0, -1).lastIndexOf('--config');
+
+    if (configPathFlagIndex !== -1) {
+      Object.assign(result, { configPath: process.argv[configPathFlagIndex + 1] });
```
yanguoyu, line 41: The config path may be set to null, is it affecting to run tasks?

packages/cli/src/main.ts
```diff
@@ -1,8 +1,100 @@
...
+    }
+
+    return result;
+  })();
+
+  const ctx = await initialKuai(args);
+  const env = ctx.getRuntimeEnvironment();
+  Object.values(env.tasks)
+    .filter((task) => !task.isSubtask)
+    .map((task) => {
```
Keith-CY, line 51: Use `forEach` instead of `map` because `map` implies the returned value is useful.

A warning will be marked by some code analysis tools
image

Same issue at, https://github.com/ckb-js/kuai/pull/32/files#diff-79567c60a701dd642531b599dee81fc4b0829920af74fd352277f88f65693527R60 https://github.com/ckb-js/kuai/pull/32/files#diff-79567c60a701dd642531b599dee81fc4b0829920af74fd352277f88f65693527R71, https://github.com/ckb-js/kuai/pull/32/files#diff-79567c60a701dd642531b599dee81fc4b0829920af74fd352277f88f65693527R88

Keith-CY (approved):

Keith-CY:
Please resolve conflicts

packages/cli/src/main.ts
PainterPuppets: No effect

PR #33, opened 2022-11-21
feat: kuai context

PainterPuppets:
create `kuaiproject/core` package & implements task system 

 - [x] create `KuaiContext` for initial kuai project

task system design: https://github.com/ckb-js/kuai/issues/13

### DIFF
--- a/packages/core/src/config/config-loading.ts
+++ b/packages/core/src/config/config-loading.ts
@@ -1,0 +1,40 @@
+import path from 'node:path';
+import { KuaiConfig, KuaiArguments } from '../type';
+import { getUserConfigPath } from '../project-structure';
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function importCsjOrEsModule(filePath: string): any {
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
+  const imported = require(filePath);
+  return imported.default !== undefined ? imported.default : imported;
+}
+
+export function resolveConfigPath(configPath: string | undefined): string | undefined {
+  if (configPath === undefined) {
+    return getUserConfigPath();
+  }
+
+  if (!path.isAbsolute(configPath)) {
+    configPath = path.join(process.cwd(), configPath);
+    configPath = path.normalize(configPath);
+  }
+
+  return configPath;
+}
+
+export async function loadConfigAndTasks(args: KuaiArguments = {}): Promise<KuaiConfig> {
+  const configPath = resolveConfigPath(args.configPath);
+
+  await import('../builtin-tasks');
+
+  let userConfig;
+
+  if (configPath) {
+    userConfig = importCsjOrEsModule(configPath);
+  }
+
+  return {
+    ...userConfig,
+    kuaiArguments: args,
+  };
+}
--- a/packages/core/src/config/config-loading.ts
+++ b/packages/core/src/config/config-loading.ts
@@ -1,6 +1,7 @@
 import path from 'node:path';
 import { KuaiConfig, KuaiArguments } from '../type';
 import { getUserConfigPath } from '../project-structure';
+import { DEFAULT_KUAI_ARGUMENTS } from '../constants';
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 function importCsjOrEsModule(filePath: string): any {
@@ -35,6 +36,10 @@ export async function loadConfigAndTasks(args: KuaiArguments = {}): Promise<Kuai
 
   return {
     ...userConfig,
-    kuaiArguments: args,
+    kuaiArguments: {
+      ...DEFAULT_KUAI_ARGUMENTS,
+      ...args,
+      configPath,
+    },
   };
 }
--- a/packages/core/src/context.ts
+++ b/packages/core/src/context.ts
@@ -1,0 +1,41 @@
+import { RuntimeEnvironment } from './type';
+import { TasksLoader } from './task-loader';
+import { ExtenderManager } from './extenders';
+import { KuaiError } from './errors';
+import { ERRORS } from './errors-list';
+
+export class KuaiContext {
+  private static __KuaiContext?: KuaiContext;
+
+  public static getInstance(): KuaiContext {
+    if (this.__KuaiContext !== undefined) {
+      return this.__KuaiContext;
+    }
+
+    const ctx = new KuaiContext();
+    this.__KuaiContext = ctx;
+    return ctx;
+  }
+
+  public deleteKuaiContext(): void {
+    KuaiContext.__KuaiContext = undefined;
+  }
+
+  public readonly tasksLoader = new TasksLoader();
+  public readonly extendersManager = new ExtenderManager();
+  public environment?: RuntimeEnvironment;
+
+  public setRuntimeEnvironment(env: RuntimeEnvironment): void {
+    if (this.environment !== undefined) {
+      throw new KuaiError(ERRORS.GENERAL.RUNTIME_ALREADY_DEFINED);
+    }
+    this.environment = env;
+  }
+
+  public getRuntimeEnvironment(): RuntimeEnvironment {
+    if (this.environment === undefined) {
+      throw new KuaiError(ERRORS.GENERAL.RUNTIME_NOT_DEFINED);
+    }
+    return this.environment;
+  }
+}
--- a/packages/core/src/project-structure.ts
+++ b/packages/core/src/project-structure.ts
@@ -1,0 +1,19 @@
+import findUp from 'find-up';
+
+const JS_CONFIG_FILENAME = 'kuai.config.js';
+const TS_CONFIG_FILENAME = 'kuai.config.ts';
+
+export function isCwdInsideProject(): boolean {
+  return findUp.sync(TS_CONFIG_FILENAME) !== null || findUp.sync(JS_CONFIG_FILENAME) !== null;
+}
+
+export function getUserConfigPath(): string | undefined {
+  const tsConfigPath = findUp.sync(TS_CONFIG_FILENAME);
+  if (tsConfigPath) {
+    return tsConfigPath;
+  }
+
+  const pathToConfigFile = findUp.sync(JS_CONFIG_FILENAME);
+
+  return pathToConfigFile;
+}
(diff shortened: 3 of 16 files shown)


Keith-CY:
There are too many conflicts that makes it hard to review, we'd better resolve them first

packages/core/src/config/config-loading.ts
```diff
@@ -0,0 +1,35 @@
+import path from 'path';
```
Keith-CY, line 1:
```suggestion
import path from 'node:path';
```

packages/core/src/config/config-loading.ts
```diff
@@ -0,0 +1,35 @@
...
+import { KuaiConfig, KuaiArguments } from '../type';
+import { getUserConfigPath } from '../project-structure';
+
+// eslint-disable-next-line
```
Keith-CY, line 5: Better to disable a specific rule

packages/core/src/config/config-loading.ts
```diff
@@ -0,0 +1,35 @@
...
+function importCsjOrEsModule(filePath: string): any {
+  // eslint-disable-next-line
+  const imported = require(filePath);
+  return imported.default !== undefined ? imported.default : imported;
+}
+
+export function resolveConfigPath(configPath: string | undefined): string | undefined {
+  if (configPath === undefined) {
+    configPath = getUserConfigPath();
```
Keith-CY, line 14: `configPath` could be returned here to avoid nested `if-else`

packages/core/src/context.ts
```diff
@@ -0,0 +1,59 @@
...
+import { ERRORS } from './errors-list';
+
+type GlobalWithKuaiContext = typeof global & {
+  __KuaiContext: KuaiContext;
+};
+
+export class KuaiContext {
+  public static isCreated(): boolean {
+    const globalWithKuaiContext = global as GlobalWithKuaiContext;
+    return globalWithKuaiContext.__KuaiContext !== undefined;
```
Keith-CY, line 14: Is it necessary to inject `__KuaiContext` into the `global` scope?

packages/core/src/project-structure.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export function getUserConfigPath(): string | undefined {
+  const tsConfigPath = findUp.sync(TS_CONFIG_FILENAME);
+  if (tsConfigPath) {
+    return tsConfigPath;
+  }
+
+  const pathToConfigFile = findUp.sync(JS_CONFIG_FILENAME);
+
+  return pathToConfigFile;
```
Keith-CY, line 18: Exception of `not found` is not handled

packages/core/src/task-loader.ts
```diff
@@ -0,0 +1,118 @@
...
+    return this._tasks;
+  }
+
+  private _addTask<ArgT extends TaskArguments>(
+    name: string,
+    descriptionOrAction?: string | ActionType<ArgT>,
+    action?: ActionType<ArgT>,
+    isSubtask?: boolean,
+  ) {
+    ``;
```
Keith-CY, line 91: What does the empty string do here

packages/core/src/project-structure.ts
PainterPuppets: Here if not found will return undefined, in the initialization of `KuaiContext` will not force must have a `kuai.config.ts`, do we want to force the user to use when there must be a config file, if not on the error?

packages/core/src/context.ts
```diff
@@ -0,0 +1,54 @@
+import { RuntimeEnvironment } from './type';
+import { TasksLoader } from './task-loader';
+import { ExtenderManager } from './extenders';
+import { KuaiError } from './errors';
+import { ERRORS } from './errors-list';
+
+export class KuaiContext {
```
yanguoyu, line 7: Maybe only `createKuaiContext` need to be set to `static`.

packages/core/src/task-loader.ts
```diff
@@ -0,0 +1,117 @@
...
+  ) {
+    const parentTask = this._tasks[name];
+
+    const task: Task = (() => {
+      if (parentTask !== undefined) {
+        return new OverrideTask(parentTask, isSubtask);
+      }
+
+      return new SimpleTask(name, isSubtask);
+    })();
```
yanguoyu, line 99:
```suggestion
    const task: Task = parentTask !== undefined ? new OverrideTask(parentTask, isSubtask) : new SimpleTask(name, isSubtask);
```

packages/core/src/context.ts
PainterPuppets: If `getKuaiContext` is not static, how should I call it, or should I remove`getKuaiContext` and use `__KuaiContext` directly?

Keith-CY: Is it possible to set and get the singleton in one action like

```typescript
getKuaiContext() {
  if (!this.__KuaiContext) {
    this.__KuaiContext = new KuaiContext()
  }
  return this.__KuaiContext
}
```

PainterPuppets: After create, some initialization work will be done (such as setting runtime) If use `getOrCreate`, there may be a problem that kuaiContext is created without initialization, which may generate some errors

Keith-CY: I didn't get the point here. The code referenced is

```typescript
export function initialKuai(args: KuaiArguments = {}): KuaiContext {
  loadTsNode();

  const ctx = KuaiContext.createKuaiContext();

  const { config } = loadConfigAndTasks(args);

  const env = new KuaiRuntimeEnvironment(config, ctx.tasksLoader.getTasks(), ctx.extendersManager.getExtenders());
  ctx.setRuntimeEnvironment(env);

  return ctx;
}
```
After expansion it would be 
```typescript
export function initialKuai(args: KuaiArguments = {}): KuaiContext {
  loadTsNode();

  const ctx = (() => {
    if (this.__KuaiContext !== undefined) {
      throw new KuaiError(ERRORS.GENERAL.CONTEXT_ALREADY_CREATED);
    }
    const _ctx = new KuaiContext();
    this.__KuaiContext = _ctx;
    return _ctx;
  })()

  const { config } = loadConfigAndTasks(args);

  const env = new KuaiRuntimeEnvironment(config, ctx.tasksLoader.getTasks(), ctx.extendersManager.getExtenders());
  ctx.setRuntimeEnvironment(env);

  return ctx;
}
```

which is exactly the same as 
```typescript
export function initialKuai(args: KuaiArguments = {}): KuaiContext {
  loadTsNode();

  const ctx = new KuaiContext()

  const { config } = loadConfigAndTasks(args);

  const env = new KuaiRuntimeEnvironment(config, ctx.tasksLoader.getTasks(), ctx.extendersManager.getExtenders());
  ctx.setRuntimeEnvironment(env);

  return ctx;
}
```

If the get and set are done in one action, the code would be expanded into the same logic here, unless the `context` has been initialized before and an exception `new KuaiError(ERRORS.GENERAL.CONTEXT_ALREADY_CREATED)` will be thrown

PainterPuppets: ok, got

Keith-CY: > In some case where just want to get the `KuaiContext` and don't know the `KuaiArguments` (usually from the cli argv / env / config file)
> 
> For example
> 
> https://github.com/ckb-js/kuai/blob/d34bdf93f7b9a1521254549cc69c0fab6f67a117/packages/core/src/config/config-env.ts#L37-L43

It's expected to throw an exception `new KuaiError(ERRORS.GENERAL.CONTEXT_NOT_CREATED)` if `initialKuai` is not done, am I right?

If yes, then I got it.

Keith-CY: But it's still confusing because the logic here is coupled. Accessibility of KuaiContext instance relies on other functions.

PainterPuppets: yes

yanguoyu: How about only letting `getKuaiContext` and `createKuaiContext` static?
If after `createKuaiContext` should init environment, why not transfer as a parameter to `createKuaiContext`?
```typescript
  public static createKuaiContext(env: RuntimeEnvironment): KuaiContext {
    if (this.__KuaiContext !== undefined) {
      throw new KuaiError(ERRORS.GENERAL.CONTEXT_ALREADY_CREATED);
    }
    const ctx = new KuaiContext();
    ctx.setRuntimeEnvironment(env)
    this.__KuaiContext = ctx;
    return ctx;
  }
```

PainterPuppets: There is nothing wrong with using `getOrCreate` here, except that if separate get and create, it is easier to find the error of not creating when using get, while using create makes the programmer aware of the need for some initialization processing here

Keith-CY: > There is nothing wrong with using `getOrCreate` here, except that if separate get and create, it is easier to find the error of not creating when using get, while using create makes the programmer aware of the need for some initialization processing here

The problem comes here.

The `getKuaiContext` is coupled with `initializeKuai` so if it's called alone, an exception `new KuaiError(ERRORS.GENERAL.CONTEXT_NOT_CREATED)` will be thrown.

However, I, as the user of this function, don't know `initializeKuai`(the message also doesn't cover it) and will only try `createKuaiContext` before `getKuaiContext`. What's worse is that `createKuaiContext` makes `getKuaiContext` work.

If it's fine to make `getKuaiContext` work by `createKuaiContext`, they would be good to be placed together.

If `createKuaiContext` is not the only condition to allow user to get the instance of `KuaiContext`, the check should be placed in the `get` method.

PainterPuppets: got it

packages/core/src/config/config-loading.ts
```diff
@@ -0,0 +1,36 @@
...
+    configPath = path.normalize(configPath);
+  }
+
+  return configPath;
+}
+
+export function loadConfigAndTasks(args: KuaiArguments = {}): { config: KuaiConfig } {
+  const configPath = resolveConfigPath(args.config);
+
+  require('../builtin-tasks');
```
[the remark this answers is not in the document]
PainterPuppets: Because `require` is synchronous and `import` is asynchronous

felicityin (approved):

yanguoyu (approved):

Issue #8, opened 2022-10-17
launch a local network for testing


homura:
## Description

As a dApp user, I want the local network to be simple enough to use out of the box.

We can take a cue from `hardhat node`

<pre>
> npx hardhat node
Started HTTP and WebSocket JSON-RPC server at http://127.0.0.1:8545/

Accounts
========

WARNING: These accounts, and their private keys, are publicly known.
Any funds sent to them on Mainnet or any other live network WILL BE LOST.

Account #0: 0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266 (10000 ETH)
Private Key: 0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80

Account #1: 0x70997970c51812dc3a010c7d01b50e0d17dc79c8 (10000 ETH)
Private Key: 0x59c6995e998f97a5a0044966f0945389dc9e86dae88c7a8412f4603b6b78690d

Account #2: 0x3c44cdddb6a900fa2b585dd299e03d12fa4293bc (10000 ETH)
Private Key: 0x5de4111afa1a4b94908f83103eb1f1706367c2e68ca870fc3fb9a804cdab365a
</pre>

1. out of box, no config required
2. 20 key pairs with assets (10000 ETH) are provided by default

Since CKB does not have a tool like Ganache (Tippy is written in C# and it is outdated) now, we can use CKB with a default config for the testing purpose as local network at first

Kuai is likely to write contracts in C/Rust in the long term, so we can't live without the RISC-V toolchain, which means we will be depending on Docker, so the local network can also depend on Docker directly, which will require Docker on the developer's machine

## Reference

- Docker [image](https://hub.docker.com/layers/nervos/ckb/merge-indexer-customize/images/sha256-62d5b54797931f02766258cc92c1e6e57cddaa689b3fbb3029c11da0ca6a4de2?context=explore)
- [dev config](https://github.com/homura/ckb-dev/blob/main/devnet/ckb/dev.toml)

PainterPuppets:
## How Hardhat uses the built-in network to help with testing

hardhat has a `HardhatNode` class, this class storage the block state, has a built-in evm to process contracts and transactions, provides some function like `mining block`, and by default, it mines a block with each transaction that it receives, in order and with no delay.

`HardhatNetworkProvider` is a provider that implements the EIP1193 interface,  and use `HardhatNode` to interact with chains

[...]

PR #37, opened 2022-11-24
feat: add docker-node packege & integrate node cli

PainterPuppets:
integrate kuai/cli & kuai/core

- [x] add `docker-node` packege
- [x] add builtin node task
- [x] node task test 

resolve: https://github.com/ckb-js/kuai/issues/8

image

image


packages/docker-node/src/docker-manager.ts
```diff
@@ -0,0 +1,162 @@
...
+      throw new DockerNotRunningError();
+    }
+
+    return new DockerManager(Docker);
+  }
+
+  public static async isInstalled(): Promise<boolean> {
+    const { exec } = await import('child_process');
+    return new Promise((resolve) => {
+      exec('which docker', (error?) => resolve(!error));
```
yanguoyu, line 33: Maybe you should use `https://github.com/shelljs/shelljs` to call `which`. Because the windows platform may not have `which` command.

packages/docker-node/src/docker-manager.ts
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  public static imageToRepoTag(image: Image): string {
+    return `${image.repository}:${image.tag}`;
+  }
+
+  readonly _docker: Docker;
+
+  // The constructor is private, see [[DockerManager.create]].
+  private constructor(DockerImpl: typeof Docker) {
```
yanguoyu, line 44: What're your means to transmit `Docker`? Why not create the `_docker` like this?
```typescript
this._docker = new Docker({ socketPath: DOCKER_SOCKET_PATH })
```

packages/docker-node/src/docker-manager.ts
```diff
@@ -0,0 +1,162 @@
...
+      throw error;
+    }
+  }
+
+  public async imageExists(image: Image): Promise<boolean> {
+    const { got } = await import('got-cjs');
+
+    const repositoryPath = this._imageToRepositoryPath(image);
+
+    const imageEndpoint = `https://registry.hub.docker.com/v2/repositories/${repositoryPath}/tags/${image.tag}/`;
```
yanguoyu, line 70: Shall we set the docker URL as an environment variable?

packages/docker-node/src/docker-manager.ts
```diff
@@ -0,0 +1,162 @@
...
+
+const DOCKER_SOCKET_PATH = '/var/run/docker.sock';
+
+export class DockerManager {
+  public static async create(): Promise<DockerManager> {
+    if (!(await DockerManager.isInstalled())) {
+      throw new DockerNotInstalledError();
+    }
+
+    if (!(await fsExtra.pathExists(DOCKER_SOCKET_PATH))) {
```
Keith-CY, line 23: [fs.existsSync(path)](https://nodejs.org/api/fs.html#fsexistssyncpath) should work here.

packages/docker-node/src/docker-manager.ts
Keith-CY, line 33: [child_process.execSync(command[, options])](https://nodejs.org/api/child_process.html#child_processexecsynccommand-options) could be used here.

packages/docker-node/src/docker-manager.ts
Keith-CY, line 70: A configurable registry would be preferable

packages/docker-node/src/docker-manager.ts
```diff
@@ -0,0 +1,162 @@
...
+      if (error instanceof DockerBadGatewayError) {
+        return false;
+      }
+
+      throw error;
+    }
+  }
+
+  public async imageExists(image: Image): Promise<boolean> {
+    const { got } = await import('got-cjs');
```
Keith-CY, line 66: Is `got-cjs` a special http client?

I ask this because another HTTP client `cross-fetch` was introduced in PR https://github.com/ckb-js/kuai/pull/37 and it has a native node module `fetch`.

Is it possible to use native `fetch` instead of the `got-cjs` package?

packages/docker-node/src/docker-manager.ts
```diff
@@ -0,0 +1,162 @@
...
+      }
+
+      if (error.statusCode === 500) {
+        throw new DockerServerError(error);
+      }
+
+      if (
+        error.statusCode === 400 &&
+        error.message !== undefined &&
+        error.message.includes('executable file not found')
```
Keith-CY, line 137:
```suggestion
        error.message?.includes('executable file not found')
```

package.json
```diff
@@ -33,9 +33,12 @@
     "typescript": "4.9.3"
   },
   "lint-staged": {
-    "**/*": [
+    "**/*.{js,ts}": [
```
Keith-CY, line 36: prettier with `--ignore-unknown` skips unsupported file types so it's safe to cover all files with prettier.

Maybe we can change the strategy to prettier all and eslint js/ts

packages/cli/src/main.ts
```diff
@@ -11,10 +10,23 @@ const KUAI_GLOBAL_PARAMS: Array<TaskParam> = [
...
 
 function parseTaskParams(param: TaskParam<TaskArguments>): Option {
-  const type = param.type.name === 'boolean' ? '' : ` <${param.type.name}>`
+  const type = (() => {
+    if (param.type.name === 'boolean') {
+      return ''
+    }
+
+    return ` <${param.type.name}>`
+  })()
```
Keith-CY, line 29: Why turns the ternary expression into an iife

packages/core/src/builtin-tasks/node.ts
```diff
@@ -0,0 +1,47 @@
...
+  .setAction(async ({ port, detached }: Args, env) => {
+    if (env.config.kuaiArguments?.network === 'docker-node') {
+      const ckbNetManager = await CkbNetManager.create()
+      ckbNetManager.start({ port: port.toString(), detached })
+
+      env.config.network = {
+        hostname: 'localhost',
+        port,
+      }
+    }
```
Keith-CY, line 25: Would it be better to throw an exception about using an unknown `network` instead of doing nothing in the else branch?

packages/core/src/builtin-tasks/node.ts
```diff
@@ -0,0 +1,47 @@
...
+  })
+
+subtask('node:stop', 'stop ckb node').setAction(async (_, env) => {
+  if (env.config.kuaiArguments?.network === 'docker-node') {
+    const ckbNetManager = await CkbNetManager.create()
+    ckbNetManager.stop()
+  }
```
Keith-CY, line 32: Would it be better to throw an exception about using an unknown `network` instead of doing nothing in the else branch?

packages/docker-node/src/docker-manager.ts
```diff
@@ -0,0 +1,161 @@
...
+      }
+
+      throw error
+    }
+  }
+
+  public async imageExists(image: Image): Promise<boolean> {
+    const repositoryPath = this._imageToRepositoryPath(image)
+
+    const imageEndpoint = `https://registry.hub.docker.com/v2/repositories/${repositoryPath}/tags/${image.tag}/`
```
Keith-CY, line 71: A configurable registry would be preferable

packages/docker-node/src/errors.ts
```diff
@@ -0,0 +1,67 @@
+import { Image } from './types'
+
+export class DockerError extends Error {
+  constructor(public readonly parent?: Error) {
+    super()
```
Keith-CY, line 5: Why only this error is instantiated with `super()` while others are instantiated with `super(parent)`

packages/docker-node/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "outDir": "lib",
+    "typeRoots": ["node_modules/@types", "node_modules/@types/node_modules/@types", "./src/types"]
```
Keith-CY, line 5: Why is `"node_modules/@types/node_modules/@types"` listed in `typeRoots`

packages/core/package.json
```diff
@@ -8,6 +8,7 @@
     "test": "jest"
   },
   "dependencies": {
-    "find-up": "5.0.0"
+    "find-up": "5.0.0",
```
Keith-CY, line 11: Is the dependency added manually? I ask this because dependencies starting with `@` should precede others

packages/cli/__tests__/index.ts
```diff
@@ -1,22 +1,112 @@
...
+    })
+
+    expect(res.status).toEqual(200)
+
+    const data = await res.json()
+    expect(data.result.objects.length).toEqual(1)
+    expect(data.result.objects[0].output.capacity).toEqual('0x1bc16d674ec80000')
+  })
+
+  test('ckb node accept multi genesis accounts', async () => {
```
yanguoyu, line 74: This test case seems the same as the above

Keith-CY:
Any update on this PR?

yanguoyu (approved):

Keith-CY (approved):

Issue #40, opened 2022-11-24
Update some configuration for development experience


Keith-CY:
1. add a `format` npm script and run it with staging files on committing;
2. update `prettier` to remove semicolons.

Keith-CY:
PR: https://github.com/ckb-js/kuai/pull/44

Keith-CY:
> Using semi will make the code style consistent, so it is recommended to use semi
> 
> > Certain ECMAScript statements (empty statement, variable statement, expression statement, do-while statement, continue statement, break statement, return statement, and throw statement) must be terminated with semicolons
> 
> [262.ecma-international.org/5.1/#sec-7.9](https://262.ecma-international.org/5.1/#sec-7.9)

This is just a matter of code style and has been agreed upon by other members.

The title of the section referenced above is `Automatic Section Insertion` which is known as `ASI` to specify how the semicolon should be inserted by **interpreter**. The conclusion of this section `These situations are described by saying that semicolons are automatically inserted into the source code token stream in those situations.` also proves it.

And the specification/TC39 committee has no bias on the code style

Ref: https://github.com/tc39/ecma262/pull/1062#pullrequestreview-330849370

PR #44, opened 2022-11-24
Update code style according to the new configuration of prettier

Keith-CY:
This PR updates the configuration of prettier to disable semicolons and update the pre-commit hook to run the `format` script instead of checking the code format.

The new style is also adopted in this PR so all unnecessary semicolons are removed.

Ref: https://github.com/ckb-js/kuai/issues/40

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -2,5 +2,6 @@
   "printWidth": 120,
   "trailingComma": "all",
   "singleQuote": true,
-  "arrowParens": "always"
+  "arrowParens": "always",
+  "semi": false
 }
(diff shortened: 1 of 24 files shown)


Keith-CY:
Most of the source codes are formatted by `prettier`, except https://github.com/ckb-js/kuai/pull/44/files#diff-3c955f0f8d911dbf3b8523f50874b91753fecdf8d4c020087761e2807e895d56R13-R16

Seems the type assertion `(Error as any).captureStackTrace` and eslint rule comment `// eslint-disable-next-line @typescript-eslint/no-explicit-any` are redundant.

yanguoyu (approved):

felicityin (approved):

Keith-CY:
Please have a review @PainterPuppets

PainterPuppets (approved):

Issue #39, opened 2022-11-24
Implement Registry of Actor model


Keith-CY:
The registry in https://github.com/ckb-js/kuai/pull/29 is roughly implemented with a `Map` for fast delivery so other developers won't be blocked by this model.

The registry is expected to be done by a container and will be implemented with [inversify](http://inversify.io/)

Keith-CY:
PR: https://github.com/ckb-js/kuai/pull/50

PR #50, opened 2022-11-30
feat: implement the registry with a container

Keith-CY:
This PR implements the registry of actor models with a container
which scans modules on initiation to bind user-defined actor classes
into the container and instantiate them on demand

Once the project starts, the registry will scan all files and bind classes
decorated by `ActorProvider` into the container and instantiate them 
lazily by calling `registry.find(uri)`

Ref: https://github.com/ckb-js/kuai/issues/39

### DIFF
--- a/packages/models/src/actor/registry.ts
+++ b/packages/models/src/actor/registry.ts
@@ -1,41 +1,70 @@
 /**
  * @tutorial https://github.com/ckb-js/kuai/issues/17#issuecomment-1305896619
- *
- * The registry is used as an IoC of actors
- * At this early stage, a custom registry is used for demostration, will be replaced by inversify
  */
+import fs from 'node:fs'
+import { resolve } from 'node:path'
 
-import { ActorRef, ActorURI } from './interface'
-import type { Actor, ActorConstructor } from './actor'
+import { Container } from 'inversify'
+import type { ActorRef, ActorURI } from './interface'
+import type { Actor } from './actor'
+import { DuplicatedActorException, InvalidActorURIException, ProviderKey } from '../utils'
 
 export class Registry {
-  #actors: Map<ActorURI, Actor> = new Map()
-  #actorConstructor: ActorConstructor
+  #actors: Set<ActorURI> = new Set()
+  #container: Container = new Container()
 
-  constructor(actorConstructor: ActorConstructor) {
-    this.#actorConstructor = actorConstructor
-  }
-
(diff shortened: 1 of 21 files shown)


felicityin (approved):

packages/models/src/actor/registry.ts
```diff
@@ -1,41 +1,70 @@
...
   }
 
   list = (): IterableIterator<ActorURI> => {
     return this.#actors.keys()
   }
 
-  isLive = (uri: ActorURI): boolean => {
-    return this.#actors.has(uri)
+  // TODO: this method is not tested
+  load = (path: string): void => {
```
yanguoyu, line 33: How about using `@Bind`, but not scanning the directory?

packages/models/src/examples/dapp/actors/child.ts
```diff
@@ -0,0 +1,5 @@
+import { ActorProvider } from '../../../'
+import { CustomActorBase } from './base'
+
+@ActorProvider({ name: 'child', path: '/parent/' })
+export class ParentActor extends CustomActorBase {}
```
yanguoyu, line 5:
```suggestion
export class ChildActor extends CustomActorBase {}
```

packages/models/src/actor/registry.ts
Keith-CY: How to detect all bound interfaces on launch without scanning the project

yanguoyu: May be like this?
```
const register = new Registry()
function Bind(target: new (...args: Array<unknown>) => unknown) {
  register.bind(target)
}

@Bind
class ChildActor {

}
```

Keith-CY: > May be like this?
> 
> ```
> const register = new Registry()
> function Bind(target: new (...args: Array<unknown>) => unknown) {
>   register.bind(target)
> }
> 
> @Bind
> class ChildActor {
> 
> }
> ```

The decorator works on instantiation so `register.bind(target)` will be evaluated on instantiation of `ChildActor`, but our goal is to instantiate `ChildActor` by the container/register. It turns into a chicken and egg problem.

PainterPuppets (approved):
LGTM

packages/models/src/actor/registry.ts
Keith-CY: I see this conversation has been marked as resolved, but this PR is not approved, are there any pending reviews not submitted?

yanguoyu (approved):

Issue #41, opened 2022-11-24
Implement of Store Modal


yanguoyu:

yanguoyu:
> Any updates on the subtasks?

https://github.com/ckb-js/kuai/pull/51
It's a draft PR.

Keith-CY:
I found that the schema defined in https://github.com/ckb-js/kuai/pull/51 missed storage of `type script` and `lock script`, correct me if I'm wrong

yanguoyu:
> I found that the schema defined in #51 missed storage of `type script` and `lock script`, correct me if I'm wrong

This is my considerations:
For a DApp application, they may only use one of `type` and `lock`, If they use more `lock` or `type`(it should include some contracts), they should define more than one schema.
So they should define some schema to handle `type` or `lock`, but not define `type` or `lock` in a schema.
But I think I lost `data` location property definition.

Keith-CY:
> > I found that the schema defined in #51 missed storage of `type script` and `lock script`, correct me if I'm wrong
> 
> This is my considerations: For a DApp application, they may only use one of `type` and `lock`, If they use more `lock` or `type`(it should include some contracts), they should define more than one schema. So they should define some schema to handle `type` or `lock`, but not define `type` or `lock` in a schema. But I think I lost `data` location property definition.

Sure, developers should define a schema for `type` or `lock` so `Store` could locate corresponding cells by `pattern`. But in current implementation, I didn't find the interface to define the schema of `type` or `lock`. Then how can we map a Store to a specific `type script`

yanguoyu:
> > > I found that the schema defined in #51 missed storage of `type script` and `lock script`, correct me if I'm wrong
> > 
> > 
> > This is my considerations: For a DApp application, they may only use one of `type` and `lock`, If they use more `lock` or `type`(it should include some contracts), they should define more than one schema. So they should define some schema to handle `type` or `lock`, but not define `type` or `lock` in a schema. But I think I lost `data` location property definition.
> 
> Sure, developers should define a schema for `type` or `lock` so `Store` could locate corresponding cells by `pattern`. But in current implementation, I didn't find the interface to define the schema of `type` or `lock`. Then how can we map a Store to a specific `type script`

Of course, I forget the `data` location property definition.

yanguoyu:
A draft [pr](https://github.com/ckb-js/kuai/pull/84) about `molecule serialize and deserialize`, I need to add `union` and `option` molecule type and add test case.
[...]

Issue #2, opened 2022-10-08
Design `Store` model


Keith-CY:
- [ ] as an actor factor: https://github.com/ckb-js/kuai/issues/4
- [ ] default storage layout
- [ ] pattern to gather cells
- [ ] interfaces
    - [ ] new(pattern): fetch cells by the specific pattern and initialize a `Store` model
    - [ ] destroy(): remove data in the `Store` model
    - [ ] duplicate(): create a new `Store` model from plain cells with the same data
    - [ ] sync(blockNumber?): load data with pattern defined on initialization at a specific block number
    - [ ] get(path): get a value at the specific path
    - [ ] set(path, value): update a value at the specific path
    - [ ] delete(path): remove key-value pair at the specific path
- [ ] metadata of a storage
	- [ ] schema
	- [ ] serialization
	- [ ] how to share/propagate metadata
- [ ] lazy evaluation of updates

Keith-CY:
A rough to-do list of `Store` model was added to this issue. Other models have their own issues
- Contract: https://github.com/ckb-js/kuai/issues/3
- Token: https://github.com/ckb-js/kuai/issues/5

We can talk a bit more about details and I'll enrich the to-do list along with the discussion. @janx @homura @felicityin @yanguoyu

yanguoyu:
Define cell storage struct to gather by the specific pattern.

yanguoyu:
Define cell storage struct to gather by the specific pattern.

Keith-CY:
> Define cell storage struct to gather by the specific pattern.

Does this mean how to match a pattern with a given cell storage struct?

Take the issues in GitHub as an example

If we are building an issue system on-chain and every issue is stored as a cell, it would be structured as
```
interface Issue {
  open: boolean
  updated_at: timestamp
  label: Array<string>
  author: User
}
```

![image](https://user-images.githubusercontent.com/7271329/194736952-85549420-c6d8-47f9-b4f5-3506fb95bde8.png)

While we can filter cells/issues out by a pattern as `is:open sort:updated-desc author:@me`

![image](https://user-images.githubusercontent.com/7271329/194736973-5039fec6-6fe7-4037-beb2-7faa2cf3bc70.png)

Would the question be `what's the default struct of a cell off-chain and corresponding patterns `?

yanguoyu:
> interface Issue {
  open: boolean
  updated_at: timestamp
  labels: Array<string>
  author: User
}
[...]

PR #51, opened 2022-11-30
feat: Implement store model

yanguoyu:
1. JSON storage
2. Pattern define
3. Store implement


PainterPuppets (approved):
Usually the serialization will have a validation solution, store will have a similar solution?

yanguoyu:
> a validation solution

Can you take an example?

PainterPuppets:
>

PainterPuppets:
> > a validation solution
> 
> Can you take an example?

sorry, point wrong

serialization of nest https://docs.nestjs.com/techniques/serialization https://github.com/typestack/class-validator

packages/models/src/contract/index.ts
```diff
@@ -1 +1,95 @@
...
+    payload: MessagePayload
+    result?: ContractMethodResponse
+  }
+}
+
+export interface State {}
+
+export default class Contract extends Store {
+
+  update(oldState: State): State {
```
Keith-CY, line 29: Would it be better to type `State` as generic

Keith-CY: BTW, this PR is specific for `Store` model. Why is the `contract` model included here

packages/models/src/store/json-storage.ts
```diff
@@ -0,0 +1,87 @@
+import { ChainStorage, StorageOnChain, StorageOffChain } from './chain-storage'
+
+type JSONStorageType = string | number | boolean | bigint
+export type JSONStorageOffChain =
+  | JSONStorageType
+  | {
+      [x: string]: JSONStorageType | JSONStorageType[] | JSONStorageOffChain
```
Keith-CY, line 7: Using `Record` makes it concise

packages/models/src/store/json-storage.ts
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+const typeMarkLen = 4
+const typeMarkMap: Record<string, string> = {
+  string: '0001',
+  number: '0010',
+  boolean: '0011',
+  bigint: '0100',
```
Keith-CY, line 15: Not sure if it's a good idea to introduce `number` and `bigint` simultaneously because precision in JavaScript leads to many troubles. How about using `big number` only?

packages/models/src/store/json-storage.ts
```diff
@@ -0,0 +1,87 @@
...
+}
+
+function replacer(_: string, value: JSONStorageOffChain | JSONStorageType) {
+  const valueType = typeof value
+  if (valueType === 'object') return value
```
Keith-CY, line 20: `null` should be elimiated

packages/models/src/store/json-storage.ts
```diff
@@ -0,0 +1,87 @@
...
+
+const typeMarkLen = 4
+const typeMarkMap: Record<string, string> = {
+  string: '0001',
+  number: '0010',
+  boolean: '0011',
+  bigint: '0100',
+}
+
+function replacer(_: string, value: JSONStorageOffChain | JSONStorageType) {
```
Keith-CY, line 18: Please add a comment about the reserved parameter, or name it meaningfully with `_` prefix.

packages/models/src/store/json-storage.ts
```diff
@@ -0,0 +1,87 @@
...
+  const valueType = typeof value
+  if (valueType === 'object') return value
+  const mark = typeMarkMap[valueType]
+  if (mark) {
+    return `${mark}${value.toString()}`
+  }
+  throw new Error('Unexpected type')
+}
+
+function reviver(_: string, value: JSONStorageOffChain | JSONStorageType) {
```
Keith-CY, line 28: Please add a comment about the reserved parameter, or name it meaningfully with `_` prefix.

packages/models/src/store/json-storage.ts
```diff
@@ -0,0 +1,87 @@
...
+  if (typeof value === 'object') {
```
Keith-CY, line 29: `null` should be eliminated

packages/models/src/store/json-storage.ts
```diff
@@ -0,0 +1,87 @@
...
+
+type JSONStorageType = string | number | boolean | bigint
+export type JSONStorageOffChain =
+  | JSONStorageType
+  | {
+      [x: string]: JSONStorageType | JSONStorageType[] | JSONStorageOffChain
+    }
+
+const typeMarkLen = 4
+const typeMarkMap: Record<string, string> = {
```
Keith-CY, line 11: Constants should be named as `TYPE_MARK_LEN` and `TYPE_MARK_MAP`

packages/models/src/store/json-storage.ts
```diff
@@ -0,0 +1,87 @@
...
+      return {
+        data: Buffer.from(JSON.stringify(data.data, replacer)),
+        witness: Buffer.from(JSON.stringify(data.witness, replacer)),
+      }
+    }
+    if ('data' in data) {
+      return {
+        data: Buffer.from(JSON.stringify(data.data, replacer)),
+      }
+    }
```
Keith-CY, line 63: A validation `'witness' in data` should be added before serialization, or an exception should be thrown.

packages/models/src/store/json-storage.ts
```diff
@@ -0,0 +1,87 @@
...
+        witness: JSON.parse(witnessStr, reviver),
+      } as T
+    }
+    if (dataStr) {
+      return {
+        data: JSON.parse(dataStr, reviver),
+      } as T
+    }
+    return {
+      witness: JSON.parse(witnessStr!, reviver),
```
Keith-CY, line 84: The assertion `witnessStr!` throws an unclear exception when `witnessStr` is `undefined`

packages/models/src/store/store.ts
```diff
@@ -0,0 +1,109 @@
+import { Actor, ActorMessage, MessagePayload } from '../actor'
+import { OutPointString, StoreMessage, StorePath } from './interface'
+import { ChainStorage, StorageOffChain } from './chain-storage'
```
Keith-CY, line 3:
```suggestion
import type { ChainStorage, StorageOffChain } from './chain-storage'
```

packages/models/src/store/store.ts
```diff
@@ -0,0 +1,109 @@
...
+import { JSONStorage, JSONStorageOffChain } from './json-storage'
+import { NonExistentException, NonStorageInstanceException } from '../exceptions'
+
+type GetState<T> = T extends ChainStorage<infer State> ? State : never
+type UnknownAsNever<T> = unknown extends T ? (0 extends 1 & T ? T : never) : T
```
Keith-CY, line 8: What does `UnknownAsNever` mean

.eslintrc.js
```diff
@@ -12,6 +12,6 @@ module.exports = {
       },
     ],
     '@typescript-eslint/no-explicit-any': 'error',
-    '@typescript-eslint/explicit-module-boundary-types': 'error',
+    '@typescript-eslint/explicit-module-boundary-types': 'off',
```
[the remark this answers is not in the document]
yanguoyu: Most of the time the return type of function can be inferenced by the statement.

packages/models/src/store/store.ts
yanguoyu: If the developer does not transfer the `data` type in `state`. They should not get `data` by `get` function.
```
get(key: OutPointString, paths: ['witness']): UnknownAsNever<GetState<StorageT>['witness']>
const store = new JSONStore<{ data: { a: number } }>()
const witness = store.get('0x1234', ['witness'])
```
The `witness` will be `never`.

packages/models/src/store/json-storage.ts
yanguoyu: `Record` will throw an error named circular reference.
image

Keith-CY: Got it

yanguoyu:
> > > a validation solution
> > 
> > 
> > Can you take an example?
> 
> sorry, point wrong
> 
> serialization of nest https://docs.nestjs.com/techniques/serialization https://github.com/typestack/class-validator

Actual, it has valid the params for `serialize` and `deserialize`, and I add some test cases for it with https://github.com/ckb-js/kuai/pull/51/commits/b51e1f66b1df1242269e681265c8af8cd51f14c8.

packages/models/src/store/chain-storage.ts
```diff
@@ -0,0 +1,12 @@
+type OneOfStorageLoc<Data, Witness = Data> = { data: Data; witness?: Witness } | { data?: Data; witness: Witness }
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type StorageOffChain<Data = any, Witness = Data> = OneOfStorageLoc<Data, Witness>
+export type StorageOnChain = OneOfStorageLoc<Uint8Array>
+
+export abstract class ChainStorage<Data extends StorageOffChain = StorageOffChain> {
+  abstract serialize(data: Data): StorageOnChain
```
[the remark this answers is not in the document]
yanguoyu: It's better to rename `Data` to `OffChain`, but `StorageOffChain` is a template type with a default parameter. It shouldn't rename to `OffchainDefault`. I add `Storage` in the prefix of `StorageOffChain` to indicate that it's storage.

packages/models/src/store/chain-storage.ts
```diff
@@ -0,0 +1,12 @@
+type OneOfStorageLoc<Data, Witness = Data> = { data: Data; witness?: Witness } | { data?: Data; witness: Witness }
```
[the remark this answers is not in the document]
yanguoyu: How about `AtLeastOneOfLoc`? I rename it to that.

packages/models/src/store/json-storage.ts
Keith-CY: But only introducing `big number` would reduce the difficulty of manipulating numbers.

If `number` and `bigint` are used in a project, their drawbacks of them revealed in JavaScript ecosystem come too. In contrast, if only `big number` is there, users don't have to think about the precision or max safe number, etc.

It may have lower performance but has greater development experience.

Keith-CY (approved):

packages/models/src/store/json-storage.ts
yanguoyu: Actually, `number`'s performance is the same as `big number`. I think it's good to remove the `number` type.

yanguoyu: https://github.com/ckb-js/kuai/pull/51/commits/3045f1f3289e1f88478c6f01cee1433811207b5d

packages/models/src/store/store.ts
```diff
@@ -0,0 +1,179 @@
...
+import { Actor } from '../actor'
+import { JSONStorage } from './json-storage'
+import { NonExistentException, NonStorageInstanceException } from '../exceptions'
+
+export class Store<
+  StorageT extends ChainStorage,
+  StorageStruct extends GetStorageStruct<GetState<StorageT>>,
+  Option = never,
+> extends Actor<StorageStruct, MessagePayload<StoreMessage<StorageStruct>>> {
+  constructor(params?: { states?: Record<OutPointString, StorageStruct>; options?: Option }) {
```
Keith-CY, line 15: I forgot the reason for introducing `outpoint` as the key of a storage struct, have we recorded it in an issue?

yanguoyu: It's used to index cells' storage data easily.

Daryl-L:
I have a question, how can I instance the `Store` object from database, by using `type script` or `lock script` or both?

Keith-CY:
> I have a question, how can I instance the `Store` object from database, by using `type script` or `lock script` or both?

`Store` is a mapping of cells specified by `pattern`, `pattern` is a combination of conditions as `SQL`(designed in https://github.com/ckb-js/kuai/issues/2#issuecomment-1305222823), e.g. `lock.code_hash eq A and lock.hash_type eq B and lock.args eq C and type.code_hash eq D and type.hash_type eq F and type.args eq G and data like '%H%'`

zhengjianhui (approved):

Keith-CY:
We will review this PR later because manipulation on `type script` and `lock script` is going to be implemented along with this feature.

Ref: https://github.com/ckb-js/kuai/issues/41#issuecomment-1366360802

yanguoyu:
`lock` and `type` schema have been added. @Keith-CY @Daryl-L @zhengjianhui @PainterPuppets

Daryl-L (approved):

Keith-CY:
Here're 2 approvals so this PR is ready to merge once conflicts are resolved

Issue #49, opened 2022-11-30
Implement of IO


PainterPuppets:
We can reference the design of https://github.com/ckb-js/kuai/pull/18

- [x] Implement `CoR`: https://github.com/ckb-js/kuai/pull/53
- [ ]  implement `Listener`: https://github.com/ckb-js/kuai/pull/66
- [ ] [TBD]Implement a Router like `koa-router`: https://github.com/ckb-js/kuai/pull/66
- [ ] implement adapter for `koa`: https://github.com/ckb-js/kuai/pull/66
- [ ] implement adapter for `express`
- [ ] implement adapter for `fastify`

PR #53, opened 2022-11-30
feat: implement io

PainterPuppets:
- [x] implement `CoR`
- [x] unit-test for `CoR` 

Resolve: https://github.com/ckb-js/kuai/issues/49

### DIFF
--- a/packages/io/src/cor.ts
+++ b/packages/io/src/cor.ts
@@ -1,0 +1,23 @@
+import { CoR as ICoR, Middleware, Context, JsonValue } from './types'
+import compose from 'koa-compose'
+
+export class CoR implements ICoR {
+  private _middlewares: Middleware[] = []
+
+  public use(plugin: Middleware): CoR {
+    this._middlewares = [...this._middlewares, plugin]
+    return this
+  }
+
+  public async dispatch<Payload extends JsonValue, Ok>(payload: Payload): Promise<Ok | void> {
+    return new Promise((resolve, rej) => {
+      const ctx: Context = {
+        payload,
+        ok: (ok?: Ok) => (ok !== undefined ? resolve(ok) : resolve()),
+        err: rej,
+      }
+
+      compose(this._middlewares)(ctx)
+    })
+  }
+}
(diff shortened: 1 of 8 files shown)


felicityin (approved):

packages/io/src/cor.ts
```diff
@@ -0,0 +1,26 @@
+import { CoR as ICoR, Middleware, Context, JsonValue } from './types'
+
+export class CoR implements ICoR {
+  private _middlewares: Middleware[] = []
+
+  public use(plugin: Middleware): void {
+    this._middlewares = [...this._middlewares, plugin]
```
Keith-CY, line 7: `this` could be returned so this method could be used as `cor.use().use().use()`

packages/io/src/cor.ts
```diff
@@ -0,0 +1,26 @@
...
+  }
+
+  public async dispatch<Payload extends JsonValue, Ok>(payload: Payload): Promise<Ok | void> {
+    return new Promise((resolve, rej) => {
+      const ctx: Context = {
+        payload,
+        ok: (ok?: Ok) => (ok ? resolve(ok) : resolve()),
```
Keith-CY, line 14: `resolve()` will be called if value of `ok` is falsy

packages/io/src/cor.ts
```diff
@@ -0,0 +1,26 @@
+import { CoR as ICoR, Middleware, Context, JsonValue } from './types'
+
+export class CoR implements ICoR {
+  private _middlewares: Middleware[] = []
+
+  public use(plugin: Middleware): void {
+    this._middlewares = [...this._middlewares, plugin]
+  }
+
+  public async dispatch<Payload extends JsonValue, Ok>(payload: Payload): Promise<Ok | void> {
```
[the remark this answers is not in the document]
Keith-CY: Maybe `compose` of `koa` could be used directly

Ref: https://github.com/koajs/compose

Keith-CY (approved):

yanguoyu (approved):

PR #66, opened 2022-12-08
feat: implement kuai-router & listener

PainterPuppets:
- [x] implement `KuaiRouter`
 - [x] implement  `KoaRouterAdapter`
 - [x] implement Listener

### DIFF
--- a/packages/io/__tests__/cor.ts
+++ b/packages/io/__tests__/cor.ts
@@ -101,4 +101,33 @@ describe('Test CoR', () => {
     expect(mockBefore).toBeCalled()
     expect(mockAfter).toBeCalled()
   })
+
+  it(`should return ok on any middleware`, async () => {
+    const cor = new CoR()
+    const mockFn = jest.fn()
+    const mockFn1 = jest.fn()
+    const mockFn2 = jest.fn()
+
+    cor.use(async (_, next) => {
+      mockFn()
+      await next()
+    })
+    cor.use(async (ctx, next) => {
+      mockFn1()
+      ctx.ok('ok')
+      await next()
+    })
+    cor.use(async (ctx, next) => {
+      mockFn2()
+      ctx.ok('hello')
+      await next()
+    })
+
+    const result = await cor.dispatch({})
+
+    expect(result).toEqual('ok')
+    expect(mockFn).toBeCalled()
+    expect(mockFn1).toBeCalled()
+    expect(mockFn2).toBeCalled()
+  })
 })
--- a/packages/io/__tests__/cor.ts
+++ b/packages/io/__tests__/cor.ts
@@ -101,32 +101,4 @@ describe('Test CoR', () => {
     expect(mockBefore).toBeCalled()
     expect(mockAfter).toBeCalled()
   })
-
-  it(`should return ok on any middleware`, async () => {
-    const cor = new CoR()
-    const mockFn = jest.fn()
-    const mockFn1 = jest.fn()
-    const mockFn2 = jest.fn()
-
-    cor.use(async (_, next) => {
-      mockFn()
-      await next()
-    })
-    cor.use(async (ctx, next) => {
-      mockFn1()
-      ctx.ok('ok')
-      await next()
-    })
-    cor.use(async (ctx) => {
-      mockFn2()
-      ctx.ok('hello')
-    })
-
-    const result = await cor.dispatch({})
-
-    expect(result).toEqual('ok')
-    expect(mockFn).toBeCalled()
-    expect(mockFn1).toBeCalled()
-    expect(mockFn2).toBeCalled()
-  })
 })
--- a/packages/io/__tests__/cor.ts
+++ b/packages/io/__tests__/cor.ts
@@ -117,10 +117,9 @@ describe('Test CoR', () => {
       ctx.ok('ok')
       await next()
     })
-    cor.use(async (ctx, next) => {
+    cor.use(async (ctx) => {
       mockFn2()
       ctx.ok('hello')
-      await next()
     })
 
     const result = await cor.dispatch({})
--- a/packages/io/__tests__/router.ts
+++ b/packages/io/__tests__/router.ts
@@ -1,0 +1,35 @@
+import { describe, it, expect } from '@jest/globals'
+import { CoR } from '../src/cor'
+import { KuaiRouter } from '../src/router'
+
+describe('test KuaiRouter', () => {
+  it(`should use kuaiRouter`, async () => {
+    const cor = new CoR()
+    const kuaiRouter = new KuaiRouter()
+    kuaiRouter.get('/', async (ctx, next) => {
+      ctx.ok('hello root')
+      await next()
+    })
+
+    kuaiRouter.get('/parent', async (ctx, next) => {
+      ctx.ok('hello parent')
+      await next()
+    })
+
+    kuaiRouter.get('/parent/children', async (ctx, next) => {
+      ctx.ok('hello children')
+      await next()
+    })
+
+    cor.use(kuaiRouter.middleware())
+
+    const rootResult = await cor.dispatch({ method: 'GET', path: '/' })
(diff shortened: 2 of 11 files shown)


packages/io/package.json
```diff
@@ -24,9 +24,14 @@
     "url": "https://github.com/ckb-js/kuai/issues"
   },
   "dependencies": {
-    "koa-compose": "4.1.0"
+    "@ckb-lumos/rpc": "0.19.0",
+    "koa": "^2.14.1",
```
Keith-CY, line 28: Dependencies should be pinned

packages/io/src/listener.ts
```diff
@@ -0,0 +1,45 @@
+import { Base } from '@ckb-lumos/rpc/lib/Base'
```
Keith-CY, line 1:
```suggestion
import type { Base } from '@ckb-lumos/rpc/lib/Base'
```

PainterPuppets:
support custom pipe

Keith-CY:
CI failed

packages/io/src/adapter.ts
```diff
@@ -0,0 +1,26 @@
+import KoaRouter from 'koa-router'
+import { CoR } from './cor'
+
+export class KoaRouterAdapter extends KoaRouter {
+  constructor(private readonly cor: CoR) {
+    super()
+  }
+
+  middleware(): KoaRouter.IMiddleware {
```
Keith-CY, line 9: What's the difference between `middleware` and `routes`

packages/io/__tests__/router.ts
```diff
@@ -0,0 +1,35 @@
...
+import { CoR } from '../src/cor'
+import { KuaiRouter } from '../src/router'
+
+describe('test KuaiRouter', () => {
+  it(`should use kuaiRouter`, async () => {
+    const cor = new CoR()
+    const kuaiRouter = new KuaiRouter()
+    kuaiRouter.get('/', async (ctx, next) => {
+      ctx.ok('hello root')
+      await next()
```
Keith-CY, line 11: `ctx.ok('hello root')` resolves the promise returned by `cor.dispatch` method, what does the following `next` do here

packages/io/src/adapter.ts
PainterPuppets: No difference, just for compatibility with koa

packages/io/__tests__/router.ts
PainterPuppets: next() really doesn't do anything, it has been removed

packages/io/__tests__/cor.ts
```diff
@@ -101,4 +101,32 @@ describe('Test CoR', () => {
...
+      mockFn2()
+      ctx.ok('hello')
+    })
+
+    const result = await cor.dispatch({})
+
+    expect(result).toEqual('ok')
+    expect(mockFn).toBeCalled()
+    expect(mockFn1).toBeCalled()
+    expect(mockFn2).toBeCalled()
```
Keith-CY, line 130: In this case, `cor` has 3 middlewares
1. call `mockFn()` and pass control to the next middleware
2. call `mockFn1()`, call `ctx.ok('ok')`, pass control to the next middleware
3. call `mockFn2()`, call `ctx.ok('hello')`, exit the control flow.

My question is about step 2. In this step `ctx.ok('ok')` is called and the promise returned by `dispatch` is resolved with `'ok'`. The workflow should finish because caller will get its response `'ok'`. Why should the control flow continue? And what would happen when `ctx.ok` is called twice?

It looks like 2 threads are working simultaneously:
1. `request | mockFn() | mockFn1() | respond`
2. `request | mockFn() | mockFn1() | mockFn2() | unknown behavior`

PainterPuppets: Additional calls to resolve() will not have any effect, the code is similar to this.

```ts
promise = new Promise(reslove => {
    reslove(1)
    // do something...
    reslove(2)
})
```
There is a discussion about "twice reslove": https://stackoverflow.com/a/29491617

Keith-CY: > Additional calls to resolve() will not have any effect, the code is similar to this.
> 
> ```ts
> promise = new Promise(reslove => {
>     reslove(1)
>     // do something...
>     reslove(2)
> })
> ```
> 
> There is a discussion about "twice reslove": [stackoverflow.com/a/29491617](https://stackoverflow.com/a/29491617)

It's safe to call `resolve` twice indeed because a promise can only be fulfilled once. But this is unexpected.

Keith-CY: To be more explicit, the workflow should finished mandatorily once `ctx.ok` is called, any updates after it is a side effect.

PainterPuppets: So how do we handle the after hook, for example if I want to know how long the request spend

``` ts
cor.use((ctx, next) => {
	const start = new Date();
	await next()
	console.log('spend', new Date() - start)
})

cor.use((ctx, next) => {
	doSomething();
	ctx.ok('ok')
})
```

This code is converted to a promise equivalent to:
``` ts
new Promise(reslove => {
	const start = new Date();
	
	doSomething();
	reslove('ok') // There is no way to handle the after hook if return directly here

	console.log('spend', new Date() - start)
})
```

Keith-CY: Let's see how `koa` handles a request

1. a request arrives and is caught by `handleRequest` method defined in `koa.callback`, in this method, koa creates a `ctx` with `req` and `res` objects, then pass `ctx` and `middlewares` into `koa.handleRequest(ctx, middlewares)` method, https://github.com/koajs/koa/blob/master/lib/application.js#L161-L164
2. In `koa.handleRequest(ctx, middlewares)` method, `ctx` is passed into `middlewares` for various pre- and post-treatments, https://github.com/koajs/koa/blob/master/lib/application.js#L188
3. the `middlewares` are composed by [koa-compose](https://github.com/koajs/compose/blob/master/index.js), in the `compose`, middlewares are called nested by `await next()`, `next()` in `middleware_i` is calling `middleware_i+1`, https://github.com/koajs/compose/blob/master/index.js#L42
4. Once the most inside middleware returns a resolved promise, all pre-treatments are done, and all post-treatments start from the inside middleware to outside middleware.
5. Each middleware resolves when its rest logic is executed and passes the control to its outside middleware, until all middlewares are resolved. The logic goes to https://github.com/koajs/koa/blob/master/lib/application.js#L188 and `ctx` is passed to `handleResponse(ctx)`, which calls `respond(ctx)`, `respond(ctx)` calls `res.end(ctx.body)` to send `response` out.

```const Koa = require('koa');
const app = new Koa();

// logger

app.use(async (ctx, next) => {
  await next();
  const rt = ctx.response.get('X-Response-Time');
  console.log(`${ctx.method} ${ctx.url} - ${rt}`);
});

// x-response-time

app.use(async (ctx, next) => {
  const start = Date.now();
  await next();
  const ms = Date.now() - start;
  ctx.set('X-Response-Time', `${ms}ms`);
});

// response

app.use(async ctx => {
  ctx.body = 'Hello World';
});

app.listen(3000);
```

is more like

```
const start = Date.now()
ctx.body = 'Hello World'
const ms = Date.now() - start
ctx.set('X-Response-Time', `${ms}ms`)
const rt = ctx.response.get('X-Response-Time')
console.log(`${ctx.method} ${ctx.url} - ${rt}`)
res.end(ctx.body)
```

Keith-CY: I think the code is fine, just the case is improper. Once `ctx.ok` is called, the control flow should be inverted from inside to outside instead of keeping moving inside

Keith-CY (approved):

zhengjianhui (approved):

yanguoyu (approved):

Keith-CY:
Is this PR ready to merge?

Daryl-L (approved):

PR #69, opened 2022-12-08
Feat contract

yanguoyu:
Add `Contract` extends `Store`, the Contract implements two functions:
1. `run` function can call remote or local service.
2. `link` function can mount functions on `Contract`, most of the time it is used to mount other's DApp's functions.


packages/models/src/contract/interface.ts
```diff
@@ -0,0 +1,12 @@
+import type { ActorURI, CallResponse, MessagePayload } from '../actor'
+
+export type Params = object
```
Keith-CY, line 3: The `Params` is so general that will conflict with one from other module, how about adding a namespace here

packages/models/src/exceptions/contract.ts
```diff
@@ -0,0 +1,5 @@
+export class ParamsMissException extends Error {
+  constructor() {
+    super(`Remote call should deliver payload and method`)
+  }
+}
```
Keith-CY, line 5: This error covers two cases, better to be explicit

packages/models/src/contract/contract.ts
```diff
@@ -0,0 +1,67 @@
...
+    if (!payload || !method) throw new ParamsMissException()
+    // call remote service
+    try {
+      const res = await request(to, {
+        method: 'POST',
+        body: JSON.stringify({
+          id: 0,
+          jsonrpc: '2.0',
+          method,
+          params: payload.value,
```
Keith-CY, line 28: Why is `method` required on a remote call but useless on a local call

Keith-CY (approved):

Keith-CY:
@Daryl-L @PainterPuppets @zhengjianhui please have a review

Daryl-L (approved):

Keith-CY:
Please squash and merge after conflicts are resolved

Issue #62, opened 2022-12-07
Implement message queue of actor model


Keith-CY:
The message queue in https://github.com/ckb-js/kuai/pull/29 is roughly implemented with a `Map` for fast delivery so other modules won't be blocked.

It's expected to be done with [redis stream](https://redis.io/docs/data-types/streams/) as designed in https://github.com/ckb-js/kuai/issues/17

Keith-CY:
It was expected to be done by 12/08, but there's no bandwidth for it. I'll try to propose a PR by 12/12

Keith-CY:
PR: https://github.com/ckb-js/kuai/pull/74

PR #74, opened 2022-12-12
feat: implement mailbox with redis stream

Keith-CY:
This PR implements the mailbox of actor model with redis stream

Ref: https://github.com/ckb-js/kuai/issues/62


PainterPuppets (approved):
lgtm

yanguoyu (approved):

Issue #85, opened 2022-12-26
Binding between Store model and cells


Keith-CY:
As the [proposal](https://github.com/ckb-js/topics/blob/add-draft-of-kuai-proposal/README.md) said
 
> Note that a Store model could be a group of cells matching the same pattern and working as an entity, so it could be regarded as a virtual DApp. Say we have 2 DApps in School Roster Store models, each of them consists of many Student Store models. And we are going to build a Sports Meeting DApp, a new Sports Meeting Store model could be created and consist of partial Student Store from School A and B, respectively, it should work as same as a School Roster DApp.

A `store` model should have a `pattern` that is used to bind to specific cells. It's defined in the [architecture](https://github.com/ckb-js/kuai/issues/7) as `Resource binding` between `CKB Node` and `Store model`.
![image](https://user-images.githubusercontent.com/7271329/209507725-4660bb8f-6b17-4de7-888c-d3b1365955e3.png)

With the binding, a `store` model maps itself to specific cells and parses data from those cells into a structured data model as the state of a dapp.

Any updates on the chain should be delivered to the `store` model and keep its state up with on-chain data.

Please add the technical design of `resource binding` between `store` and `ckb node` in this issue, and then we can make a plan for the implementation.

Keith-CY:
Could you take over this issue as a starting point for joining this project? @zhengjianhui @Daryl-L 
Any question could be appended in this issue.

Daryl-L:
> Could you take over this issue as a starting point for joining this project? @zhengjianhui @Daryl-L Any question could be appended in this issue.

No problem~

Daryl-L:
From my perspective, in the [IO Design](https://github.com/ckb-js/kuai/issues/15), `resource binding` module could subscribe to the block change from IO component, and then fetch the data on chain to publish to the `Store` model.

Keith-CY:
> From my perspective, in the [IO Design](https://github.com/ckb-js/kuai/issues/15), `resource binding` module could subscribe to the block change from IO component, and then fetch the data on chain to publish to the `Store` model.

It should work, but more details are required, like how to fetch data on chain by the pattern defined in `Store`

Daryl-L:
From this issue #2, each `Store` may have multiple states,  and each state is related to one cell. So does it mean that the relationship between `Store` and cell is many-to-many?

Keith-CY:
[...]

PR #94, opened 2023-01-05
feat(io): export the modules in @kuai/io

Daryl-L:
I think the listener in `@kuai/io` should be exported, then the `resource binding` in issue #85 could use it.

### DIFF
--- a/packages/io/package.json
+++ b/packages/io/package.json
@@ -4,7 +4,9 @@
   "description": "kuai io layer",
   "homepage": "https://github.com/ckb-js/kuai#readme",
   "license": "MIT",
-  "main": "lib/index.js",
+  "exports": {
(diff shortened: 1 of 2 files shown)


PainterPuppets (approved):
LGTM

packages/io/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "kuai io layer",
   "homepage": "https://github.com/ckb-js/kuai#readme",
   "license": "MIT",
-  "main": "lib/io.js",
```
Keith-CY, line 7: The `exports` field is recommended prior to the `main` fields

ref: https://nodejs.org/api/packages.html#package-entry-points

Keith-CY (approved):

zhengjianhui (approved):

PR #99, opened 2023-01-08
feat: Add handle cell to deserialize

yanguoyu:
Add new type for `handleCall`, reference https://github.com/ckb-js/kuai/issues/85#issuecomment-1371862113
Because of need to handle [Cell](https://github.com/ckb-js/lumos/blob/b63350a4970e6e49707e631944f349635dbfb9fc/packages/base/src/api.ts#L81), so add some functions to serialize from off-chain data and deserialize from on-chain data


Keith-CY:
Please add a brief description of this PR and related issue so reviewers can get the background of these updates.

packages/models/src/store/interface.ts
```diff
@@ -73,29 +72,58 @@ type GetScriptOption<T extends ScriptSchema> = IsEmptyScriptSchemaNever<
...
+
 export type StorageLocation = 'data' | 'witness' | ['lock' | 'type', keyof ScriptSchema]
 
 export type StorePath<K = StorageLocation> = K extends string[] ? [...K, ...string[]] : [K, ...string[]]
 
-export interface StoreMessage<State> {
-  type: 'add_state' | 'remove_state'
-  add?: Record<OutPointString, State>
-  remove?: OutPointString[]
+export type UpdateStorageValue = {
```
Keith-CY, line 116: How can I update `type` and `lock` of a storage?

yanguoyu: They can update by the `set` function.

Keith-CY: Got it

packages/models/__tests__/store/store.ts
```diff
@@ -40,125 +60,304 @@ Reflect.defineMetadata(ProviderKey.Actor, { ref }, JSONStore)
...
+        const expected = bytes.hexify(new JSONStorage().serialize({ a: BigNumber(20) }))
+        expect(chainData.witness).toBe(expected)
+        expect(chainData.data).toBeUndefined()
+        expect(chainData.lock).toBeUndefined()
+        expect(chainData.type).toBeUndefined()
+      })
+      it('init with lock', () => {
+        const store = new JSONStore<{ lock: { args: BigNumber; codeHash: { schema: { a: string }; offset: 10 } } }>({
+          lock: { args: true, codeHash: { offset: 10 } },
+        })
```
Keith-CY, line 116: A question not related to function. is the schema defined reversed, code hash is expected to be a hash string while args could be deserialized to a structure, in actual use cases.

packages/models/src/store/store.ts
```diff
@@ -10,57 +12,129 @@ import type {
...
+    ) {
+      return [option.offset || 0, option.length || 0]
+    }
+    if ('offset' in option && typeof option.offset === 'number') {
+      return [option.offset || 0, 0]
+    }
+    if ('length' in option && typeof option.length === 'number') {
+      return [0, option.length || 0]
+    }
+    return [0, 0]
```
Keith-CY, line 80: It's prone to miss the branch. I would suggest getting `offset` and `length` first, e.g.

```
const offset = 'offset' in option && typeof option.offset === 'number' ? option.offset : 0
const length = 'length' in option && typeof option.length === 'number' ? option.length : 0
return [offset, length]
```

packages/models/src/store/store.ts
```diff
@@ -10,57 +12,129 @@ import type {
...
+  }
+
+  private deserializeScript<T extends 'lock' | 'type'>(
+    scriptValue: Script,
+    scriptOption: unknown,
+    type: T,
+  ): GetFullStorageStruct<StructSchema>[T] {
+    if (typeof scriptOption !== 'object' || scriptOption === null) throw new Error()
```
Keith-CY, line 88: Error message is expected to be added

packages/models/src/store/store.ts
```diff
@@ -10,57 +12,129 @@ import type {
...
-  private addState(addStates: Record<OutPointString, GetStorageStruct<StructSchema>>) {
-    this.states = {
-      ...this.states,
-      ...addStates,
+  private addState({ cell, witness }: UpdateStorageValue) {
+    if (cell.outPoint) {
+      const outpoint = `${cell.outPoint?.txHash}${cell.outPoint.index.slice(2)}`
+      this.states[outpoint] = this.deserializeCell({ cell, witness })
+      this.chainData[outpoint] = { cell, witness }
     }
```
Keith-CY, line 131: Expect to throw if `cell` doesn't have its out point

packages/models/__tests__/store/store.ts
yanguoyu: Do you mean like this? https://github.com/yanguoyu/kuai/blob/feat-add-type/packages/models/__tests__/store/store.ts#L216-L257

packages/models/src/store/store.ts
```diff
@@ -72,22 +146,184 @@ export class Store<
...
+    type: 'lock' | 'type',
+    key: OutPointString,
+    newValue: GetFullStorageStruct<StructSchema>['lock'] | GetFullStorageStruct<StructSchema>['type'],
+  ) {
+    if ('args' in newValue) {
+      this.updateValueInCell([type, 'args'], key, newValue.args)
+    }
+    if ('codeHash' in newValue) {
+      this.updateValueInCell([type, 'codeHash'], key, newValue.codeHash)
+    }
```
Keith-CY, line 209: `hashType` should be configurable too, if `code hash` is writable.
At least they should be paired because `code hash + hash type` works together as a pointer to a script.

packages/models/src/store/store.ts
```diff
@@ -72,22 +146,184 @@ export class Store<
...
+      }
+    })
+    return clone
+  }
+
+  initOnChain(value: GetStorageStruct<StructSchema>): GetOnChainStorage<StructSchema> {
+    const res: StorageSchema<string> = {}
+    if ('data' in value) {
+      const { offset, hexString } = this.serializeField('data', value.data)
+      res.data = `0x${offset ? '0'.repeat(offset) : ''}${hexString.slice(2)}`
```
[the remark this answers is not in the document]
yanguoyu: Of course, I forget to consider it. If someone set offset or length, it should be byte length. I will resolve it after an update.

packages/models/__tests__/store/store.ts
Keith-CY: Yes, `args` consists of multiple fields in some cases, but `code hash` is definitely a hash string from `cell.typeScript` or `cell.data`.

packages/models/src/store/store.ts
```diff
@@ -72,22 +146,184 @@ export class Store<
...
+        cellInfo.cell.data =
+          cellInfo.cell.data.slice(0, 2 + offset) +
+          hexString.slice(2) +
+          (length ? cellInfo.cell.data.slice(2 + offset + length) : '')
+      } else {
+        cellInfo.witness =
+          cellInfo.witness.slice(0, 2 + offset) +
+          hexString.slice(2) +
+          (length ? cellInfo.witness.slice(2 + offset + length) : '')
+      }
```
Keith-CY, line 227: I would suggest checking `type` explicitly, and throwing an exception if `type` is neither `data` nor `type`.

It's redundant but ease for mental burden

packages/models/__tests__/store/store.ts
yanguoyu: If `args` include multiple fields, developers can define multiple schema to create multiple `store`

packages/models/src/store/store.ts
yanguoyu: I thought no added is enough.

packages/models/src/store/store.ts
yanguoyu: There is an update for off-chain data, I thought `hashType` will not store off-chain data.

packages/models/src/store/store.ts
yanguoyu: I add `type === 'witness'` with `else if`

packages/models/src/store/interface.ts
```diff
@@ -71,31 +71,60 @@ type PickExist<T, K extends keyof T> = OmitByValue<{
...
 }
+
+export type StoreMessage =
+  | {
+      type: 'remove_cell'
+      value: OutPointString[]
+    }
+  | {
+      type: 'update_cell'
+      value: UpdateStorageValue | UpdateStorageValue[]
```
Keith-CY, line 129: I would suggest defining 
```
{ type: 'update_cells`, value: UpdateStorageValue[] }
```
Because it covers `UpdateStorageValue | UpdateStorageValue[]` and `update_cells` is clearly for multiple cells.

packages/models/__tests__/store/store.ts
Keith-CY: The point is that it's anti-intuitive to impose a scheme on code hash in CKB because code hash has a clear definition

> The hash of ELF formatted RISC-V binary that contains a CKB script. For space efficiency consideration, the actual script is attached to current transaction as a dep cell. Depending on the value of hash_type, the hash specified here should either match the hash of cell data part in the dep cell, or the hash of type script in the dep cell. The actual binary is loaded into an CKB-VM instance when they are specified upon the transaction verification.

It is a string from an unidirectional algorithm.

Ref: https://github.com/nervosnetwork/rfcs/blob/master/rfcs/0019-data-structures/0019-data-structures.md#description-1

packages/models/src/store/store.ts
Keith-CY: I didn't get the point of `update for off-chain data`

Say that we have a `store` which is a group of cells with `lock script`
```
code_hash: A
hash_type: type
args: B
```
And a request is to change their lock algorithm to
```
code_hash: C
hash_type: data
args: B
```
How can we achieve that?

yanguoyu: I got that, so I need to add a function to update `code_hash` or `hash_type`.
This function is an updated side-effect by [set](https://github.com/yanguoyu/kuai/blob/aa6cbf53aae3f599c496865df1b76d18ea36a888/packages/models/src/store/store.ts#L420)

packages/models/__tests__/store/store.ts
yanguoyu: Got it, I will remove `code_hash` from `schema`.

packages/models/src/store/store.ts
yanguoyu: But the `store` is designed to manage `schema data`, should we need to manage `cell` data?

packages/models/__tests__/store/store.ts
yanguoyu: If so, the `lock` will only include `args`, I think I can use `lockArgs` for the schema defined.

packages/models/src/store/store.ts
Keith-CY: > But the `store` is designed to manage `schema data`, should we need to manage `cell` data?

`Store` is an abstraction layer to read and write data of cells. Data is a general concept, including every field of a cell, `type script`, `lock script`, `capacity`, `output data`, and `witness`.

Meanwhile, `type script` and `lock script` have their own schema, in the schema, `hash type` and `code hash` have pre-defined schema `hash type: type | data`, `code hash: string` and `args` has a flexible schema which should be defined by users.

yanguoyu: But actually, `hash_type` or `code_hash` are meaningless to `serialize` or `deserialize`.
So I defined schema like this
```
{
  data?: FieldSchema<T>
  witness?: FieldSchema<T>
  lockArgs?: FieldSchema<T>
  typeArgs?: FieldSchema<T>
}
```

Keith-CY: > But actually, `hash_type` or `code_hash` are meaningless to `serialize` or `deserialize`. So I defined schema like this
> 
> ```
> {
>   data?: FieldSchema<T>
>   witness?: FieldSchema<T>
>   lockArgs?: FieldSchema<T>
>   typeArgs?: FieldSchema<T>
> }
> ```

I think it's clear, `code hash` and `hash type` could be updated by `lockArgs.codeHash =` and `lockArgs.hashType =` while `args` could be updated by custom schema as `lockArgs.args.flag =`

yanguoyu: Yeah, `data`, `witness`, `lockArgs`, `typeArgs ` can be updated or saved by the `set` function or `update_cells` message. And the `set` function has a side effect on `cell`.
If they need to update `hash_type` or `code_hash`, I think I can add an update function to update these.

Keith-CY: > Yeah, `data`, `witness`, `lockArgs`, `typeArgs ` can be updated or saved by the `set` function or `update_cells` message. And the `set` function has a side effect on `cell`. If they need to update `hash_type` or `code_hash`, I think I can add an update function to update these.

Better to add an `update hash type` option because updating `code hash` has its own, back to the topic,
`hash type` should be accessible in `updateScriptInCell`(https://github.com/yanguoyu/kuai/blob/aa6cbf53aae3f599c496865df1b76d18ea36a888/packages/models/src/store/store.ts#L192-L203) because `code hash` requires it to be changed in some cases.

yanguoyu: https://github.com/yanguoyu/kuai/blob/aa6cbf53aae3f599c496865df1b76d18ea36a888/packages/models/src/store/store.ts#L192-L203 is called when developers call `set` to update `state`, so I need to sync these changes to `cell`.
But the `schema`s don't include `hash_type`, so when developers call `set` to update `state`, it doesn't need to update `hash_type`, I think I should expose a public function to update `hash_type` or `code_hash` only.
And after I update the schema definition to
```
{
  data?: FieldSchema<T>
  witness?: FieldSchema<T>
  lockArgs?: FieldSchema<T>
  typeArgs?: FieldSchema<T>
}
```
I have deleted the `updateScriptInCell` function.

Keith-CY (approved):

Daryl-L (approved):

Keith-CY:
There're 2 approvals and no more unresolved conversations, could we merge this PR?

yanguoyu:
> There're 2 approvals and no more unresolved conversations, could we merge this PR?

I think it's ok

Issue #96, opened 2023-01-06
Implement of resource binding component


Daryl-L:
We can reference the design of #85 

- [x] Message define
- [x] Manager define
- [x] tests

Daryl-L:
pr #100

PR #100, opened 2023-01-08
feat: implement resource binding component

Daryl-L:
1. Message implementation
2. Manager implementation

Related to issue #96


packages/models/src/resource-binding/manager.ts
```diff
@@ -0,0 +1,168 @@
...
+  listen(pollingInterval = 1000): { subscription: Subscription; updator: NodeJS.Timer } {
+    return { subscription: this._listener.on(this.onListenBlock), updator: this.update(pollingInterval) }
+  }
+
+  private removeFromInput(input: Input) {
+    const outpoint = outpointToOutPointString(input.previousOutput)
+    const store = this.#registryOutpoint.get(outpoint)
+    if (store) {
+      this.call(store.uri, {
+        pattern: 'normal',
```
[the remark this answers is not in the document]
Daryl-L: I copied from the test case in your pr #51. I don't know what value should I fill in this property, could you tell me about this?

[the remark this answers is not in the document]
Daryl-L: The pattern will be set when the store is registered. And only register message has this property.

yanguoyu (approved):

packages/models/src/resource-binding/manager.ts
```diff
@@ -0,0 +1,168 @@
...
+
+  private update(pollingInterval = 1000) {
+    return setInterval(async () => {
+      if (this.#topBlockNumber.gt(0) && this.#topBlockNumber.gt(BI.from(this.#lastBlock?.header.number ?? 0))) {
+        this.updateStore(await this._dataSource.getBlock(this.#topBlockNumber.toHexString()))
+      }
+    }, pollingInterval)
+  }
+
+  private updateStore(block: Block) {
```
Keith-CY, line 38: What if update cells in a batch, as

1. get all inputs in the block
2. get all outputs in the block
3. filter outputs that consumed in the block
4. filter inputs that consume an output in the block
5. remove remaining inputs and add remaining outputs

packages/models/src/resource-binding/manager.ts
```diff
@@ -0,0 +1,168 @@
...
+import { OutPointString } from '..'
+import type { Subscription } from 'rxjs'
+import { ChainSource } from '@kuai/io/lib/types'
+
+export class Manager extends Actor<object, MessagePayload<ResourceBindingManagerMessage>> {
+  #registry: Map<TypescriptHash, Map<LockscriptHash, ResourceBindingRegistry>> = new Map()
+  #registryOutpoint: Map<OutPointString, ResourceBindingRegistry> = new Map()
+  #registryReverse: Map<ActorURI, [TypescriptHash, LockscriptHash]> = new Map()
+  #lastBlock: Block | undefined = undefined
+  #topBlockNumber = BI.from(0)
```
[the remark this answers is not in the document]
Daryl-L: > Is `top block number` the `tip block number` of CKB node?
> 
> Ref: https://github.com/nervosnetwork/ckb/tree/develop/rpc#method-get_tip_block_number

Yes, wrong name here.

packages/models/src/resource-binding/manager.ts
Daryl-L: > What if update cells in a batch, as
> 
> 1. get all inputs in the block
> 2. get all outputs in the block
> 3. filter outputs that consumed in the block
> 4. filter inputs that consume an output in the block
> 5. remove remaining inputs and add remaining outputs

I just don't well understand. Does it mean to filter the input cells and output cells by the `Store`s and fill them in one message for each instance of `Store`?

packages/models/src/resource-binding/utils.ts
```diff
@@ -0,0 +1,7 @@
+import { OutPointString } from '../store'
+import { OutPoint } from '@ckb-lumos/base'
+import { BI } from '@ckb-lumos/bi'
+
+export function outpointToOutPointString(outpoint: OutPoint): OutPointString {
+  return `${outpoint.txHash}${BI.from(outpoint.index).toHexString().substring(2)}`
```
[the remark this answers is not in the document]
Daryl-L: > `index` wouldn't be greater than `Number.MAX_SAFE_INTEGER`, so `BI` is unnecessary.

The outpoint is assembled by myself in manager and I use the number directly, so I use the BI to format the number to HexString. I will modify the code to format the number first, and the code here will be maintained.

packages/models/src/resource-binding/manager.ts
Daryl-L: > 

OK, I see, this is better. I will improve it by this way.

zhengjianhui (approved):

packages/models/src/resource-binding/manager.ts
Daryl-L: @yanguoyu To achieve this, I think the `Store` may be added a new message payload to support to update outputs and remove inputs together. Maybe like this:

```typescript
interface Message {
  type: 'update_cells',
  update: {
    witness: HexString,
    cell: Cell
  }[],
  remove: OutPointString[]
}
```

Is that ok?

yanguoyu: I think we can send two types of messages like this.
```typescript
export type StoreMessage =
  | {
      type: 'remove_cell'
      value: OutPointString[]
    }
  | {
      type: 'update_cell'
      value: UpdateStorageValue | UpdateStorageValue[]
    }

```

Daryl-L: I have made manager to send two messages to each store when block came, one for update outputs and one for remove inputs. 

Tests about this will be committed today.

Daryl-L: > > > What if update cells in a batch, as
> > > 
> > > 1. get all inputs in the block
> > > 2. get all outputs in the block
> > > 3. filter outputs that consumed in the block
> > > 4. filter inputs that consume an output in the block
> > > 5. remove remaining inputs and add remaining outputs
> > 
> > 
> > I just don't well understand. Do you mean to filter the input cells and output cells by the `Store`s and fill them in one message for each instance of `Store`?
> 
> Similarly, the logic in the PR is to handle cells one by one from the perspective of a transaction, it would write/update the state multiple times. What if we consider the cells affected from the perspective of a block, some cells are destroyed and some cells are generated in the block.
> 
> Once cells are inspected from the block's level, the most brilliant point is that many `internal changes` in a block could be ignored, e.g., sequential transactions.
> 
> in block #n
> 
> 1. Cell_A -> | tx_1 | -> Cell_B
> 2. Cell_B -> | tx_2 | -> Cell_C
> 3. Cell_C -> | tx_3 | -> Cell_D
> 
> Only Cell_A is eliminated and Cell_D is created.
> 
> An analogy in front-end is the reconciliation in react, multiple actions would take place in a frame, but only the last state will be rendered.
> 
> `Update cells` instead of `update a cell` is of course a good improvement.

I have a simple question about this.
Could tx1 and tx2 be in the same block? If so, I think that Cell_B has not been generated onchain while sending the tx2. Because the Cell_B will be generated when in tx1, which will be executed in the same block with tx2.

Keith-CY: The example shows transactions in one block. Although `Cell_B` is consumed immediately, it still has a record on the chain.

A similar technology called `lightning network` discards the intermediate status and only the `start status` and `end status` are submitted on the chain.

Ref: https://lightning.network/

packages/models/src/resource-binding/utils.ts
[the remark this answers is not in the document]
Keith-CY: `(+outpoint.index).toString(16)` should work, `BI` is consuming.

packages/models/src/resource-binding/manager.ts
zhengjianhui: > The example shows transactions in one block. Although `Cell_B` is consumed immediately, it still has a record on the chain.
> 
> A similar technology called `lightning network` discards the intermediate status and only the `start status` and `end status` are submitted on the chain.

My understanding is that Cell_A and Cell_B can't be in the same tx, but they can be in the same block, i.e. the output of tx1 of a block can be the input of tx2, right?

yanguoyu: I'm sorry to tell you it will change to this
```typescript
{ type: 'update_cells`, value: UpdateStorageValue[] }
```

packages/models/src/resource-binding/manager.ts
```diff
@@ -0,0 +1,201 @@
...
+        )
+      }
+      this.#lastBlock = block
+    }
+  }
+
+  private filterOutputs(tx: Transaction, block: Block): Map<OutPointString, [Cell, string]> {
+    const outputs = new Map<OutPointString, [Cell, string]>()
+    for (const outputIndex in tx.outputs) {
+      if (tx.hash) {
```
Keith-CY, line 65: The `if(tx.hash)` condition could be moved outside the `for-loop`

packages/models/src/resource-binding/manager.ts
```diff
@@ -0,0 +1,201 @@
...
+          store,
+          'remove_cell',
+          input.map((v) => v.previousOutput),
+        )
+      }
+      this.#lastBlock = block
+    }
+  }
+
+  private filterOutputs(tx: Transaction, block: Block): Map<OutPointString, [Cell, string]> {
```
Keith-CY, line 62: What does the `filter` mean in this method, the logic seems to be an iteration to return all outputs of a transaction

packages/models/src/resource-binding/manager.ts
```diff
@@ -0,0 +1,201 @@
...
+        outputs.set(outPointToOutPointString(outPoint), [
+          {
+            cellOutput: tx.outputs[outputIndex],
+            data: tx.outputsData[outputIndex],
+            outPoint,
+            blockHash: block.header.hash,
+            blockNumber: block.header.number,
+          },
+          tx.witnesses[outputIndex],
+        ])
```
Keith-CY, line 79: What if merge `witness` into the first element of the tuple?

packages/models/src/resource-binding/manager.ts
```diff
@@ -0,0 +1,201 @@
...
+import { ResourceBindingRegistry, ResourceBindingManagerMessage } from './interface'
+import { outPointToOutPointString } from './utils'
+import { Listener } from '@ckb-js/kuai-io'
+import type { Subscription } from 'rxjs'
+import { ChainSource } from '@ckb-js/kuai-io/lib/types'
+import { OutPointString } from '../store'
+
+export class Manager extends Actor<object, MessagePayload<ResourceBindingManagerMessage>> {
+  #registry: Map<TypeScriptHash, Map<LockScriptHash, ResourceBindingRegistry>> = new Map()
+  #registryOutpoint: Map<OutPointString, ResourceBindingRegistry> = new Map()
```
Keith-CY, line 14: `#registryOutpoint` should be `#registerOutPoint` for `out point` in the camel case style

packages/models/src/resource-binding/manager.ts
```diff
@@ -0,0 +1,201 @@
...
+    let outputs = new Map<OutPointString, [Cell, string]>()
+    for (const tx of block.transactions) {
+      outputs = new Map([...outputs, ...this.filterOutputs(tx, block)])
+      for (const input of tx.inputs) {
+        const outPointString = outPointToOutPointString(input.previousOutput)
+        if (outputs.has(outPointString)) {
+          outputs.delete(outPointString)
+        } else {
+          const registry = this.#registryOutpoint.get(outPointString)
+          if (registry) {
```
Keith-CY, line 97: Please add the `else-branch` with `// ignore` etc. to imply that the `else` branch is neglectable by design

packages/models/src/resource-binding/manager.ts
Daryl-L: I will change the name to "mapOutputs"

packages/models/src/resource-binding/manager.ts
Daryl-L: > What if merge `witness` into the first element of the tuple?

I use the `Cell` interface which is defined in lumos to represent the first element of the tuple , it does not contain the `witnesses` property, so I put it as the second element of the tuple.

Keith-CY: Got it

packages/models/src/resource-binding/manager.ts
Daryl-L: This `else-branch` is necessary, because if the outpoint of the `input` is not in the filtered `output`s, it must be sent to the `Store` before.

Keith-CY: Seems that you're talking about the `else branch` of `if statement` at line 93 https://github.com/ckb-js/kuai/pull/100/files#diff-6beced91eeabe7f90f6b66349a7a50e5a2e7cac2c73c536c6c32131c8082ddcdR93

The one I asked about is of `if statement` line 99 https://github.com/ckb-js/kuai/pull/100/files#diff-6beced91eeabe7f90f6b66349a7a50e5a2e7cac2c73c536c6c32131c8082ddcdR99, it doesn't have an `else branch,` but I was not clear why it could be ignored. So I would like it to be annotated.

Daryl-L: > Seems that you're talking about the `else branch` of `if statement` at line 93 https://github.com/ckb-js/kuai/pull/100/files#diff-6beced91eeabe7f90f6b66349a7a50e5a2e7cac2c73c536c6c32131c8082ddcdR93
> 
> The one I asked about is of `if statement` line 99 https://github.com/ckb-js/kuai/pull/100/files#diff-6beced91eeabe7f90f6b66349a7a50e5a2e7cac2c73c536c6c32131c8082ddcdR99, it doesn't have an `else branch,` but I was not clear why it could be ignored. So I would like it to be annotated.

From my understanding, an input cell, if the current block doesn't output it, it must be output by a former block. If it isn't registered by any `Store`s, then it certainly belongs to any other applications.

Keith-CY: > > Seems that you're talking about the `else branch` of `if statement` at line 93 [#100 (files)](https://github.com/ckb-js/kuai/pull/100/files#diff-6beced91eeabe7f90f6b66349a7a50e5a2e7cac2c73c536c6c32131c8082ddcdR93)
> > The one I asked about is of `if statement` line 99 [#100 (files)](https://github.com/ckb-js/kuai/pull/100/files#diff-6beced91eeabe7f90f6b66349a7a50e5a2e7cac2c73c536c6c32131c8082ddcdR99), it doesn't have an `else branch,` but I was not clear why it could be ignored. So I would like it to be annotated.
> 
> From my understanding, an input cell, if the current block doesn't output it, it must be output by a former block. If it isn't registered by any `Store`s, then it certainly belongs to any other applications.

Got it

Keith-CY (approved):

Keith-CY:
There're 3 approvals and no more unresolved conversations, could we merge this PR?

PR #103, opened 2023-01-10
feat(@kuai/io): support params

PainterPuppets:
- `KuaiRouter` support parse params payload
 - `KoaRouterAdapter` support pass body &  query payload

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -32,6 +32,7 @@
       "resolved": "https://registry.npmjs.org/@ampproject/remapping/-/remapping-2.2.0.tgz",
       "integrity": "sha512-qRmjj8nj9qmLTQXXmaR1cck3UXSRMPrbsLJAasZpF+t3riI71BXed5ebIOYwQntykeZuhjsdweEc9BxH5Jc26w==",
       "dev": true,
+      "license": "Apache-2.0",
       "dependencies": {
         "@jridgewell/gen-mapping": "^0.1.0",
         "@jridgewell/trace-mapping": "^0.3.9"
@@ -45,6 +46,7 @@
       "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.18.6.tgz",
       "integrity": "sha512-TDCmlK5eOvH+eH7cdAFlNXeVJqWIQ7gW9tY1GJIpUtFb6CmjVyq2VM3u71bOyR8CRihcCgMUYoDNyLXao3+70Q==",
       "dev": true,
+      "license": "MIT",
       "dependencies": {
         "@babel/highlight": "^7.18.6"
       },
@@ -57,6 +59,7 @@
       "resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.20.1.tgz",
       "integrity": "sha512-EWZ4mE2diW3QALKvDMiXnbZpRvlj+nayZ112nK93SnhqOtpdsbVD4W+2tEoT3YNBAG9RBR0ISY758ZkOgsn6pQ==",
       "dev": true,
+      "license": "MIT",
       "engines": {
         "node": ">=6.9.0"
       }
@@ -66,6 +69,7 @@
       "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.20.2.tgz",
       "integrity": "sha512-w7DbG8DtMrJcFOi4VrLm+8QM4az8Mo+PuLBKLp2zrYRCow8W/f9xiXm5sN53C8HksCyDQwCKha9JiDoIyPjT2g==",
       "dev": true,
+      "license": "MIT",
       "dependencies": {
         "@ampproject/remapping": "^2.1.0",
         "@babel/code-frame": "^7.18.6",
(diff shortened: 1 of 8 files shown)


Keith-CY:
Conflicted, and please rebase on the develop branch to make the CI pass

package-lock.json
```diff
@@ -2983,32 +2982,37 @@
...
-      "dev": true,
       "dependencies": {
         "@types/node": "*"
       }
     },
     "node_modules/@types/content-disposition": {
       "version": "0.5.5",
       "resolved": "https://registry.npmjs.org/@types/content-disposition/-/content-disposition-0.5.5.tgz",
-      "integrity": "sha512-v6LCdKfK6BwcqMo+wYW05rLS12S0ZO0Fl4w1h4aaZMD7bqT3gVUns6FvLJKGZHQmYn3SX55JWGpziwJRwVgutA==",
-      "dev": true
```
Keith-CY, line 3005: `"dev": true` was removed, it's similar to this case https://github.com/npm/cli/issues/5128

I'm using npm@8.19.2, which is delivered with node@18, and run `rm package-lock.json && npx lerna bootstrap`, the `dev` field exists

packages/io/src/adapter.ts
```diff
@@ -15,10 +16,9 @@ export class KoaRouterAdapter extends KoaRouter {
       ctx.body = await this.cor.dispatch({
         method: ctx.method,
         path: ctx.path,
-        params: ctx.params,
-        // todo: support body & query
-        // query: ctx.query,
-        // body: ctx.request.body,
+        query: ctx.query as JsonValue,
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        body: (ctx.request as any)?.body || undefined,
```
Keith-CY, line 21: Is the type definition of `request` unpredictable?

packages/io/src/types/index.ts
```diff
@@ -1,29 +1,35 @@
 import type { CKBComponents } from '@ckb-lumos/rpc/lib/types/api'
-
+import { Key } from 'path-to-regexp'
```
Keith-CY, line 2:
```suggestion
import type { Key } from 'path-to-regexp'
```

packages/io/src/types/index.ts
```diff
@@ -1,29 +1,35 @@
...
+}
 
-export interface Context {
-  payload: JsonValue
+export type JsonValue = null | number | string | { [key: string]: JsonValue } | JsonValue[]
 
-  ok(): void
+type Next = () => Awaited<void>
 
+export type Context<CustomT = DefaultContext> = {
```
Keith-CY, line 19: Seems that `CustomT = DefaultContext` could be `CustomT extends DefaultContext`

packages/io/src/types/index.ts
```diff
@@ -1,29 +1,35 @@
...
 }
 
-export type Middleware = (ctx: Context, next: () => Promise<void>) => Promise<void>
+export interface DefaultContext {
+  /**
+   * Custom properties.
+   */
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  [key: string]: any
+}
```
Keith-CY, line 13: What's the difference between `DefaultContext` and `object`

packages/io/src/router.ts
```diff
@@ -9,50 +10,63 @@ export function isRoute(x: unknown): x is Route {
 }
 
 function matchPath(path: Path, route: Route): boolean {
-  // todo: add more match rule
-  return path === route.path
+  return route.regexp.test(path)
+}
+
+function makeRoute(options: { path: Path; middleware: Middleware<RouterContext>; method: Method }): Route {
```
Keith-CY, line 16: Function name `makeRoute` is hard to understand.

It's an internal function, so we can name it a long but detailed text

packages/io/src/types/index.ts
PainterPuppets: Actually, no difference

Keith-CY:
Any update on this issue?

PainterPuppets:
> Any update on this issue?

There is a type of problem is still being solved, should be almost done

packages/io/src/types/index.ts
```diff
@@ -1,29 +1,35 @@
...
 
-export interface Context {
-  payload: JsonValue
+export type JsonValue = null | number | string | { [key: string]: JsonValue } | JsonValue[]
 
-  ok(): void
+type Next = () => Awaited<void>
```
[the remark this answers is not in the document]
PainterPuppets: Awaited can be`await`, usually the middleware will be a promise function

package-lock.json
Keith-CY: `dev: true` is still removed, which version of `npm` is used on your machine

packages/io/src/adapter.ts
```diff
@@ -15,10 +24,9 @@ export class KoaRouterAdapter extends KoaRouter {
       ctx.body = await this.cor.dispatch({
         method: ctx.method,
         path: ctx.path,
-        params: ctx.params,
-        // todo: support body & query
-        // query: ctx.query,
-        // body: ctx.request.body,
+        query: ctx.query as JsonValue,
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        body: ctx.request.body || undefined,
```
Keith-CY, line 29: Is falsy value disallowed to be assigned on `body`, so add `undefined` as a fallback?

packages/io/src/types/index.ts
```diff
@@ -1,31 +1,34 @@
 import type { CKBComponents } from '@ckb-lumos/rpc/lib/types/api'
-
+import { Key } from 'path-to-regexp'
```
Keith-CY, line 2:
```suggestion
import type { Key } from 'path-to-regexp'
```

packages/io/src/types/index.ts
```diff
@@ -1,31 +1,34 @@
...
 export interface Listener<T> {
   on(listen: (obj: T) => void): void
 }
 
-export type Middleware = (ctx: Context, next: () => Promise<void>) => Promise<void>
+export type JsonValue = null | number | string | { [key: string]: JsonValue } | JsonValue[]
```
Keith-CY, line 7: Could `{ [key: string]: JsonValue }` be expressed with `Record<string, JsonValue>`?

PainterPuppets: use`Record<string, JsonValue> ` will have `Type alias 'JsonValue' looply references itself` error

Keith-CY: Got it

package-lock.json
PainterPuppets: I've been using global lerna before
After changing it to "npx lerna" it works.

packages/io/src/adapter.ts
PainterPuppets: No, I just thought there would be no body when get method, so I added an undefined, but after thinking about it, I don't really need it.

Keith-CY (approved):

Keith-CY:
One more approval is required to merge

Daryl-L (approved):

PR #121, opened 2023-01-31
feat(@kuai/core): project template integrate `kuai/io` & `kuai/models`

PainterPuppets:
- Project template integrated with `kuai/io` and `kuai/models`
- Added a  built-in hello-world controller 
- Added automatic download of dependencies at project initial

## How to start in local:
 1. Compile kuai: run `lerna run build` in the `kuai` project root directory
 2. Link the local packages: `npm link ./packages/core ./packages/cli ./packages/models ./packages/io`
 3. Switch to the project directory and run `kuai init`.

### DIFF
--- a/packages/core/sample-projects/src/type-extends.ts
+++ b/packages/core/sample-projects/src/type-extends.ts
@@ -1,0 +1,7 @@
+import '@ckb-js/kuai-core';
+
+declare module '@ckb-js/kuai-core' {
+  export interface KuaiConfig {
+    port?: number;
+  }
+}
--- a/packages/core/recommended-package.json
+++ b/packages/core/recommended-package.json
@@ -1,3 +1,7 @@
 {
-  "name": "kuai-project"
+  "name": "kuai-project",
+  "scripts": {
+    "build": "tsc",
+    "test": "jest"
+  }
 }
--- a/packages/core/sample-projects/README.md
+++ b/packages/core/sample-projects/README.md
@@ -1,3 +1,11 @@
 # Sample Kuai Project
 
 This project demonstrates a basic kuai use case.
+
+## Run server
+
+```
+npm run build
+
+node ./dist/src/main.js
+```
--- a/packages/core/sample-projects/kuai.config.ts
+++ b/packages/core/sample-projects/kuai.config.ts
@@ -1,5 +1,7 @@
 import { KuaiConfig } from '@ckb-js/kuai-core';
 
-const config: KuaiConfig = {};
+const config: KuaiConfig = {
+  port: 3000,
+};
 
 export default config;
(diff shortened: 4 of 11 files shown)


packages/core/sample-projects/src/type-extends.ts
```diff
@@ -0,0 +1,7 @@
+import '@ckb-js/kuai-core';
+
+declare module '@ckb-js/kuai-core' {
+  export interface KuaiConfig {
```
Keith-CY, line 4: What if extend KuaiConfig in https://github.com/ckb-js/kuai/pull/121/files#diff-e3c4f2238ce558dfb4f70f29fa93d80948da48dafbceba387770fff0176793aeR9 directly

Keith-CY (approved):

packages/core/sample-projects/src/type-extends.ts
PainterPuppets: The config field type of `KuaiRuntimeEnvironment` is fixed, so if directly extend `kuaiConfig` directly here will be a bit of a problem https://github.com/ckb-js/kuai/pull/121/files#diff- 12405526a43f467c07f021e225857683dcfdfb8088135c5a5ab575a578ac8f49R10-R11

If you think redeclare is not good, I will add a todo first and then change it to generate type

Keith-CY: > The config field type of `KuaiRuntimeEnvironment` is fixed, so if directly extend `kuaiConfig` directly here will be a bit of a problem [#121 (files)](https://github.com/ckb-js/kuai/pull/121/files#diff-12405526a43f467c07f021e225857683dcfdfb8088135c5a5ab575a578ac8f49R10-R11)
> 
> If you think redeclare is not good, I will add a todo first and then change it to generate type

Marking it `TODO` is fine, let's finish the MVP first

Daryl-L (commented):
I think there will be plenty of sample projects, so the structor of directory maybe like this.

```
|--- sample-projects
  |--- mvp1
    |--- src
      |--- controllers
      |--- views
      |--- models
  |--- mvp2
```

PainterPuppets:
> I think there will be plenty of sample projects, so the structor of directory maybe like this.
> 
> ```
> |--- sample-projects
>   |--- mvp1
>     |--- src
>       |--- controllers
>       |--- views
>       |--- models
>   |--- mvp2
> ```

A little bit didn't get
Meaning we will have multiple back-end servers?

Keith-CY:
> > I think there will be plenty of sample projects, so the structor of directory maybe like this.
> > ```
> > |--- sample-projects
> >   |--- mvp1
> >     |--- src
> >       |--- controllers
> >       |--- views
> >       |--- models
> >   |--- mvp2
> > ```
> 
> A little bit didn't get Meaning we will have multiple back-end servers?

There could be many `sample projects` as https://github.com/nestjs/nest/tree/master/sample, but one template is enough for now.

Daryl-L (approved):

PR #124, opened 2023-02-02
feat: create mvp dapp sample

PainterPuppets:
create mvp dapp by `kuai init`


Daryl-L (approved):

packages/samples/mvp-dapp/.prettierrc
```diff
@@ -0,0 +1,6 @@
+{
+  "printWidth": 120,
+  "trailingComma": "all",
+  "singleQuote": true,
+  "arrowParens": "always"
```
Keith-CY, line 5: A point was neglected before that it's better to keep the rule consistent: https://github.com/ckb-js/kuai/blob/develop/.prettierrc#L6

packages/samples/mvp-dapp/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "kuai-mvp-dapp",
+  "scripts": {
+    "build": "tsc"
+  },
+  "dependencies": {
+    "@ckb-js/kuai-core": "0.0.1",
+    "@ckb-js/kuai-io": "0.0.1",
+    "@ckb-js/kuai-models": "0.0.1",
+    "koa": "^2.14.1",
```
Keith-CY, line 10: Dependencies should be pinned

Keith-CY:
It seems that codes are quite different from what we've drafted in https://github.com/Magickbase/kuai-mvp-server/pull/1/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R8

Keith-CY:
Skeleton code in `controller` was simplified

yanguoyu:
After finishing MVP-DApp, we need to perfect the init project template.

Keith-CY (approved):

PR #125, opened 2023-02-03
Feature/resource binding initiate store

Daryl-L:
1. Create a `Buffer` to store cells while the `Store` is not initiated.
2. Initiate the `Store` while registered.


packages/io/__tests__/listener.ts
```diff
@@ -5,6 +5,7 @@ import { timeout, distinctUntilChanged, catchError } from 'rxjs'
 import type { Subscription } from 'rxjs'
 
 import { ChainSource } from '../src/types'
+import { Script } from '@ckb-lumos/base'
```
Keith-CY, line 8:
```suggestion
import type { Script } from '@ckb-lumos/base'
```

packages/io/__tests__/listener.ts
```diff
@@ -51,6 +52,12 @@ describe('test Listener', () => {
       }),
     getCurrentEpoch: () => Promise.resolve(mockEpoch),
     getBlock: () => Promise.resolve(mockBlock),
+    getAllLiveCellsWithWitness: function (
```
Keith-CY, line 55: Set `_lockScript` prior to `_typeScript` if `_typeScript` is optional, then the method could be called as `getAllLiveCellsWithWitness(lock)` instead of `getAllLiveCellsWithWitness(, lock)` if `type script` is `undefined`.

And `_typeScript: Script | undefined` could be declared as `_typeScript?: Script`

packages/io/src/types/index.ts
```diff
@@ -1,5 +1,7 @@
 import type { CKBComponents } from '@ckb-lumos/rpc/lib/types/api'
 import type { Key } from 'path-to-regexp'
+import { Script } from '@ckb-lumos/base'
```
Keith-CY, line 3:
```suggestion
import type { Script } from '@ckb-lumos/base'
```

packages/models/src/resource-binding/cell-change-buffer.ts
```diff
@@ -0,0 +1,56 @@
+import { Input } from '@ckb-lumos/base'
+import { ActorURI } from '../actor'
+import { ResourceBindingRegistry } from './interface'
+import { CellChangeData } from './types'
```
Keith-CY, line 4:
```suggestion
import type { Input } from '@ckb-lumos/base'
import type { ActorURI } from '../actor'
import type { ResourceBindingRegistry } from './interface'
import type { CellChangeData } from './types'
```

packages/models/src/resource-binding/cell-change-buffer.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export class CellChangeBuffer {
+  #buffer: Map<ActorURI, [ResourceBindingRegistry, Input[], CellChangeData[]][]> = new Map()
```
Keith-CY, line 7: `[ResourceBindingRegistry, Input[], CellChangeData[]]` is complex and reused in several places, it's better to assign it to a custom type.

packages/models/src/resource-binding/cell-change-buffer.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  hasReadyStore(): boolean {
+    return this.#readyList.length > 0
+  }
+
+  push(uri: ActorURI, data: [ResourceBindingRegistry, Input[], CellChangeData[]]) {
+    let buffer = this.#buffer.get(uri)
+    if (!buffer) {
+      buffer = []
+    }
```
Keith-CY, line 18:
```suggestion
    let buffer = this.#buffer.get(uri) ?? []
```

packages/models/src/resource-binding/cell-change-buffer.ts
Keith-CY: Seems that it's a `bufferMap` instead of a `buffer`?

I ask this because it is exported as `buffer` in https://github.com/ckb-js/kuai/pull/125/files#diff-4d312df7669a766201d2e72e4db825d5779742390b5e6290f189381535f02eb8R49-R51
And buffer is expected to be a data structure providing `push` and `pop`, but in this case the `buffer` doesn't have these methods.

packages/models/src/resource-binding/cell-change-buffer.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  popAll(): [ResourceBindingRegistry, Input[], CellChangeData[]][][] {
+    let length = this.#readyList.length
+    const res: [ResourceBindingRegistry, Input[], CellChangeData[]][][] = []
+    while (length-- > 0) {
+      const data = this.pop()
+      if (data) {
+        res.push(data)
+      }
+    }
```
Keith-CY, line 40: Use `this.#readyList.forEach` instead of a `while loop`

Ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach

packages/models/src/resource-binding/types.ts
```diff
@@ -1,4 +1,8 @@
-import { Hash } from '@ckb-lumos/base'
+import { Cell, Hash } from '@ckb-lumos/base'
 
 export type TypeScriptHash = Hash | 'null'
 export type LockScriptHash = Hash
+
+export type RegistryStatus = 'registered' | 'initiated'
+
+export type CellChangeData = [Cell, string]
```
Keith-CY, line 8: Use **labeled tuple** for readability

Ref: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-0.html#labeled-tuple-elements

packages/models/src/resource-binding/manager.ts
```diff
@@ -28,43 +30,65 @@ export class Manager extends Actor<object, MessagePayload<ResourceBindingManager
   }
 
   private update(pollingInterval = 1000) {
+    let semaphore = 0
     return setInterval(async () => {
-      if (this.#tipBlockNumber.gt(0) && this.#tipBlockNumber.gt(BI.from(this.#lastBlock?.header.number ?? 0))) {
-        this.updateStore(await this._dataSource.getBlock(this.#tipBlockNumber.toHexString()))
+      semaphore++
+      if (semaphore == 1) {
```
Keith-CY, line 36:
```suggestion
      if (semaphore === 1) {
```

packages/models/src/resource-binding/manager.ts
```diff
@@ -28,43 +30,65 @@ export class Manager extends Actor<object, MessagePayload<ResourceBindingManager
...
+      this.sendMessage(
+        store,
+        'remove_cell',
+        input.map((v) => outPointToOutPointString(v.previousOutput)),
+      )
+    }
+  }
+
+  private updateStore(block: Block) {
+    if (!this.#lastBlock || BI.from(block.header.number).gt(BI.from(this.#lastBlock.header.number))) {
```
Keith-CY, line 81:
```suggestion
    if (BI.from(block.header.number).gte(BI.from(this.#lastBlock?.header.number ?? 0))) {
```

Daryl-L: I think `if (!this.#lastBlock || BI.from(block.header.number).gt(BI.from(this.#lastBlock.header.number))) {` is more readable than `if (BI.from(block.header.number).gte(BI.from(this.#lastBlock?.header.number ?? 0))) {`.

packages/models/src/resource-binding/cell-change-buffer.ts
Keith-CY: > > Use `this.#readyList.forEach` instead of a `while loop`
> > Ref: [developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach)
> 
> `Store`s are popped from ready list one by one, so I use `while loop` instead of `forEach`. Another reason is I want to process a certain number of `Store` in ready list each round.

I didn't get the point because `forEach` handles elements one by one, too. And `forEach` could iterate a certain number by **array slice**.

packages/models/src/resource-binding/manager.ts
```diff
@@ -28,43 +36,65 @@ export class Manager extends Actor<object, MessagePayload<ResourceBindingManager
...
     return setInterval(async () => {
-      if (this.#tipBlockNumber.gt(0) && this.#tipBlockNumber.gt(BI.from(this.#lastBlock?.header.number ?? 0))) {
-        this.updateStore(await this._dataSource.getBlock(this.#tipBlockNumber.toHexString()))
+      semaphore++
+      if (semaphore === 1) {
+        if (this.#buffer.hasReadyStore()) {
+          this.updateBuffer()
+        }
+        if (this.#tipBlockNumber.gt(0) && this.#tipBlockNumber.gt(BI.from(this.#lastBlock?.header.number ?? 0))) {
+          this.updateStore(await this._dataSource.getBlock(this.#tipBlockNumber.toHexString()))
```
Keith-CY, line 47: Will `await this._dataSource.getBlock(this.#tipBlockNumber.toHexString())` be `undefined` or throw an error?

Keith-CY (approved):

packages/models/src/resource-binding/manager.ts
```diff
@@ -28,43 +36,72 @@ export class Manager extends Actor<object, MessagePayload<ResourceBindingManager
   }
 
   private update(pollingInterval = 1000) {
+    let semaphore = 0
```
yanguoyu, line 39: Maybe it can define as a `boolean` type?

yanguoyu:
I can't find the logic that when hash has registered but `store` does not init finish, and we need to save the cells in the buffer.
May you point out these for me?

Daryl-L:
> I can't find the logic that when hash has registered but `store` does not init finish, and we need to save the cells in the buffer. May you point out these for me?

Thanks for reminding me, I left it in stash list on my computer.

packages/models/src/resource-binding/manager.ts
Daryl-L: > Maybe it can define as a `boolean` type?

The `semaphore` might be larger than 1, `update` will be processed only when the `semaphore` is equal to 1.

yanguoyu: How about like this?
```
let semaphore = false 
return setInterval(async () => {
      if (!semaphore) {
        semaphore = true
        // do something
        semaphore = false
      }
    }, pollingInterval)
```

Daryl-L: Suppose there are two processes `A` and `B`.
When `A` read the `semaphore`, it is `false`. At the same time, `B` read the `semaphore`, and the semaphore has not  been set to `true` yet by `A`, it is still `false`. Because `if (!semaphore) {` and `semaphore = true` are not atomic.
I think it is possible.

yanguoyu: Got it, but the semaphore is not shared for two processes. And for one process it will execute in sequence.

packages/models/src/resource-binding/manager.ts
```diff
@@ -28,43 +36,79 @@ export class Manager extends Actor<object, MessagePayload<ResourceBindingManager
...
-        if (input.length > 0) {
-          this.sendMessage(
-            store,
-            'remove_cell',
-            input.map((v) => outPointToOutPointString(v.previousOutput)),
-          )
+  private updateStore(block: Block) {
+    if (!this.#lastBlock || BI.from(block.header.number).gt(BI.from(this.#lastBlock.header.number))) {
+      const changes = this.filterCellsAndMapChanges(block)
+      for (const [store, input, data] of changes) {
```
Daryl-L, line 96: > I can't find the logic that when hash has registered but `store` does not init finish, and we need to save the cells in the buffer. May you point out these for me?

@yanguoyu here

yanguoyu (approved):

PR #126, opened 2023-02-04
feat: Define Record model

yanguoyu:


packages/samples/mvp-dapp/src/record/record.model.ts
```diff
@@ -0,0 +1,32 @@
...
+  label: string
+}
+
+export type StoreType = {
+  data: {
+    profile: ItemData[]
+    addresses: ItemData[]
+    custom: ItemData[]
+    dweb: ItemData[]
+    permissions: ItemData[]
```
Keith-CY, line 15:
```suggestion
```

`permissions` could be removed because it's derived from the lock script

Keith-CY (approved):

Daryl-L (approved):

PR #127, opened 2023-02-04
feat(sample): view for sample

Daryl-L:


packages/samples/mvp-dapp/src/views/tx.view.ts
```diff
@@ -0,0 +1,26 @@
+import { CellChangeData } from '@ckb-js/kuai-models'
+import { Cell, helpers, commons } from '@ckb-lumos/lumos'
+
+export class Tx {
+  static async toJson(
+    txSkeleton: helpers.TransactionSkeletonType,
```
Keith-CY, line 6: Is the `skeleton` available for unlocking the omnilock?

packages/samples/mvp-dapp/src/views/read.view.ts
```diff
@@ -0,0 +1,13 @@
+// import type { ItemData } from '../record/record.model'
+
```
Keith-CY, line 2:
```suggestion
```

packages/samples/mvp-dapp/src/views/load.view.ts
```diff
@@ -0,0 +1,60 @@
+// import type { StoreType } from '../record/record.model'
+
```
Keith-CY, line 2:
```suggestion
```

packages/samples/mvp-dapp/src/views/load.view.ts
```diff
@@ -0,0 +1,60 @@
...
+    return JSON.stringify({
+      profile,
+      addresses,
+      custom,
+      dweb,
+    })
+  }
+}
+
+interface Item {
```
Keith-CY, line 57: Declaration `interface Item` could be placed with `type ItemData` and `type StoreType`

packages/samples/mvp-dapp/src/views/load.view.ts
```diff
@@ -0,0 +1,60 @@
...
+  data: {
+    profile: ItemData[]
+    addresses: ItemData[]
+    custom: ItemData[]
+    dweb: ItemData[]
+  }
+}
+
+export class Load {
+  static toJson(storage: StoreType): string {
```
Keith-CY, line 19: The method name indicates a `JSON` is returned, but a string is returned in fact.

packages/samples/mvp-dapp/src/views/read.view.ts
Daryl-L: Will be uncommented after model branch merged.

packages/samples/mvp-dapp/src/views/load.view.ts
Daryl-L: Will be uncommented after model branch merged.

packages/samples/mvp-dapp/src/views/read.view.ts
Keith-CY: Gotcha

packages/samples/mvp-dapp/src/views/load.view.ts
Daryl-L: Just refer to this https://github.com/Magickbase/kuai-mvp-dapp-ui/blob/develop/utils/mock.ts

Keith-CY: The point is that `type declaration` could be placed together, and should be placed before where it's used.

yanguoyu:
I think `ctx.ok`'s first parameter can be an object, why do we need to transfer it to a string?

Daryl-L:
> I think `ctx.ok`'s first parameter can be an object, why do we need to transfer it to a string?

I just referred to what @PainterPuppets did.

packages/samples/mvp-dapp/src/views/tx.view.ts
Keith-CY: Any update on this question?

Daryl-L: Supported, in https://github.com/ckb-js/kuai/pull/127/commits/cdb1aad571e0b66fe51ebd850896d4b5016b3a97.

yanguoyu (approved):

packages/samples/mvp-dapp/src/views/tx.view.ts
```diff
@@ -0,0 +1,45 @@
...
+    txSkeleton.update('outputs', (outputs) => {
+      return outputs.push(...outputCells)
+    })
+
+    for (const [input, witness] of inputData) {
+      txSkeleton.update('inputs', (inputs) => inputs.push(input))
+
+      txSkeleton.update('witnesses', (witnesses) => {
+        if (witness === '0x' || witness === '') {
+          const omniLock = config.getConfig().SCRIPTS.OMNI_LOCK as NonNullable<config.ScriptConfig>
```
Keith-CY, line 22: Seems that chain type `mainnet/testnet` is not set, but we can do it later.

Keith-CY (approved):

PR #137, opened 2023-02-09
feat: Omnilock Model and Record Model

yanguoyu:
1. Add Omnilock Model
2. Add update and clear in Record Model
3. Add process in app controller


packages/io/src/types/index.ts
```diff
@@ -57,11 +57,10 @@ export type RoutePayload<
   Body extends object = object,
 > = {
   query?: Query
-  body?: Body
-  params?: Params
   path: Path
   method: Method
-}
+} & ([Params] extends [never] ? Record<string, never> : { params: Params }) &
```
Keith-CY, line 62: AFAIK, `never` should be used as a `response` of a function only. In this case, what does `Record<string, never>` mean

packages/models/src/store/pattern-match.ts
```diff
@@ -0,0 +1,16 @@
...
+  }
+  return pattern(value, true)
+}
+
+export function schemaPatternMatch(pattern: SchemaPattern, value: unknown) {
+  if (Array.isArray(pattern)) {
+    return pattern.reduce((result, func) => result && func(value, result), true)
+  }
+  return pattern(value, true)
+}
```
Keith-CY, line 16: The logic seems almost the same in these two functions

packages/models/src/store/store.ts
```diff
@@ -76,6 +86,8 @@ export class Store<
   }
 
   private deserializeCell({ cell, witness }: UpdateStorageValue): GetStorageStruct<StructSchema> {
+    if (!this.schemaOption || typeof this.schemaOption !== 'object' || this.schemaOption == null)
```
Keith-CY, line 89: `this.schemaOption == null` is redundant because `typeof this.schemaOption !== 'object'` filter `null` out

packages/io/src/types/index.ts
yanguoyu: `Record<string, never>` means `{}`

Keith-CY: Cool, got new knowledge

Keith-CY (approved):

packages/samples/mvp-dapp/src/app.controller.ts
```diff
@@ -1,95 +1,124 @@
...
+  let omnilockModel = appRegistry.find<OmnilockModel>(actorRef.uri)
+  if (!omnilockModel) {
+    class NewStore extends OmnilockModel {}
+    Reflect.defineMetadata(ProviderKey.Actor, { ref: actorRef }, NewStore)
+    Reflect.defineMetadata(ProviderKey.CellPattern, createCellPattern(lock), NewStore)
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    appRegistry.bind(NewStore as any)
+    omnilockModel = appRegistry.find<OmnilockModel>(actorRef.uri)
+    if (!omnilockModel) throw new Error('ominilock bind error')
+    resourceBinding.register(lock, undefined, actorRef.uri, lockHash)
```
Daryl-L, line 45: I think maybe it is better to interact with resource binding manager by message bus.

Daryl-L (approved):

PR #138, opened 2023-02-10
Feature/resource binding init sample

Daryl-L:
1. add implementation of `chain source`
2. add bootstrap of `resource binding`

### DIFF
--- a/packages/samples/mvp-dapp/kuai.config.js
+++ b/packages/samples/mvp-dapp/kuai.config.js
@@ -1,5 +1,4 @@
 module.exports = {
   port: 3000,
-  rpcUrl: '',
-  indexerUrl: '',
+  rpcUrl: 'https://mainnet.ckbapp.dev/rpc',
 }
--- a/packages/samples/mvp-dapp/kuai.config.js
+++ b/packages/samples/mvp-dapp/kuai.config.js
@@ -1,3 +1,5 @@
 module.exports = {
   port: 3000,
+  rpcUrl: '',
+  indexerUrl: '',
 }
--- a/packages/samples/mvp-dapp/src/chain-source.ts
+++ b/packages/samples/mvp-dapp/src/chain-source.ts
@@ -1,6 +1,6 @@
 import type { ChainSource } from '@ckb-js/kuai-io/lib/types'
 import { Script, RPC, BI } from '@ckb-lumos/lumos'
-import { CKBComponents } from '@ckb-lumos/rpc/lib/types/api'
+import type { CKBComponents } from '@ckb-lumos/rpc/lib/types/api'
 
 export class NervosChainSource implements ChainSource {
   #rpc: RPC
--- a/packages/samples/mvp-dapp/src/chain-source.ts
+++ b/packages/samples/mvp-dapp/src/chain-source.ts
@@ -5,7 +5,7 @@ import { CKBComponents } from '@ckb-lumos/rpc/lib/types/api'
 export class NervosChainSource implements ChainSource {
   #rpc: RPC
 
-  constructor(rpcUrl: string) {
+  constructor(rpcUrl: string, private _getCellsLimit: number = 1000) {
     this.#rpc = new RPC(rpcUrl)
   }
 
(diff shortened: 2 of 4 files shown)


packages/samples/mvp-dapp/src/main.ts
```diff
@@ -1,15 +1,35 @@
...
+  Reflect.defineMetadata(
+    ProviderKey.Actor,
+    {
+      ref: {
+        name: 'resource',
+        path: '',
+        protocal: 'local',
+        uri: 'local://resource',
+      },
+    },
```
yanguoyu, line 24:
```suggestion
    new ActorReference('resource',  '')
```

packages/samples/mvp-dapp/src/main.ts
```diff
@@ -1,15 +1,35 @@
...
+    Manager,
+  )
+
+  const dataSource = new NervosChainSource(kuaiEnv.config.rpcUrl)
+  const listener = new TipHeaderListener(dataSource)
+  const manager = new Manager(listener, dataSource)
+  manager.listen()
```
yanguoyu, line 31: The `manager` should export for other modules.

packages/samples/mvp-dapp/src/chain-source.ts
```diff
@@ -0,0 +1,58 @@
+import { ChainSource } from '@ckb-js/kuai-io/lib/types'
```
Keith-CY, line 1:
```suggestion
import type { ChainSource } from '@ckb-js/kuai-io/lib/types'
```

packages/samples/mvp-dapp/src/chain-source.ts
```diff
@@ -0,0 +1,58 @@
...
+import { Script, RPC, BI } from '@ckb-lumos/lumos'
+import { CKBComponents } from '@ckb-lumos/rpc/lib/types/api'
```
Keith-CY, line 3:
```suggestion
import type { CKBComponents } from '@ckb-lumos/rpc/lib/types/api'
```

packages/samples/mvp-dapp/kuai.config.js
```diff
@@ -1,3 +1,5 @@
 module.exports = {
   port: 3000,
+  rpcUrl: '',
+  indexerUrl: '',
```
Keith-CY, line 4: `indexerUrl` is not used

packages/samples/mvp-dapp/src/main.ts
Daryl-L: What I design is to interact with resource binding manager by message bus.

yanguoyu: Will it bind in the `Register`? Will exist some manager instances or only one instance in the DApp process?

Daryl-L: > Will it bind in the `Register`? Will exist some manager instances or only one instance in the DApp process?

It will bind in `Register`, and I think there will be only one manager instance in a DApp.

yanguoyu (approved):

Keith-CY:
Conflicted

Keith-CY:
> Conflicted

Conflicts should be resolved by https://github.com/Daryl-L/kuai/pull/1

Keith-CY:
I've tried running on this branch and find the mvp-dapp doesn't work

1. Path of `resource actor` should be the root, `/` or `undefined` are both fine, but explicit `''` makes it empty. ref: https://github.com/ckb-js/kuai/pull/138/files#diff-38d98fc038c1a05b8da2d5c2c9e0e9db88c3ed2eb646dbe4334e9f861eaa865cR15
2. After setting the path of `resource actor` to `/`, the app still breaks down

Daryl-L:
> I've tried running on this branch and find the mvp-dapp doesn't work
> 
> 1. Path of `resource actor` should be the root, `/` or `undefined` are both fine, but explicit `''` makes it empty. ref: https://github.com/ckb-js/kuai/pull/138/files#diff-38d98fc038c1a05b8da2d5c2c9e0e9db88c3ed2eb646dbe4334e9f861eaa865cR15
> 2. After setting the path of `resource actor` to `/`, the app still breaks down

resolved

zhengjianhui (approved):

PR #145, opened 2023-02-15
feat: mvp dapp support dotenv

PainterPuppets:
- mvp dapp support dotenv

### DIFF
--- a/packages/samples/mvp-dapp/kuai.config.js
+++ b/packages/samples/mvp-dapp/kuai.config.js
@@ -1,5 +1,8 @@
+require('dotenv').config()
+
 module.exports = {
-  port: 3000,
-  rpcUrl: 'https://testnet.ckb.dev/rpc',
-  lumosConfig: 'aggron4',
+  host: process.env.HOST,
+  port: process.env.PORT,
+  rpcUrl: process.env.CKB_RPC_URL || 'https://testnet.ckb.dev/rpc',
+  lumosConfig: process.env.LUMOS_CONFIG || 'aggron4',
 }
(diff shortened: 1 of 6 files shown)


packages/samples/mvp-dapp/.env.sample
```diff
@@ -0,0 +1,3 @@
+ENV=dev
```
Keith-CY, line 1: Usually, `example` is preferable to `sample` since they are slightly different

Keith-CY: And the `ENV` is not used in the project

packages/samples/mvp-dapp/src/main.ts
```diff
@@ -42,7 +42,7 @@ async function bootstrap() {
         return _address
       }
 
-      return `http://${_address.address}:${_address.port}`
+      return `http://localhost:${_address.port}`
```
Keith-CY, line 45: `localhost` will be mapped to IPv6 by default when the app is running with Node.js 18

packages/samples/mvp-dapp/kuai.config.js
```diff
@@ -1,4 +1,6 @@
+require('dotenv').config()
+
 module.exports = {
-  port: 3000,
-  rpcUrl: 'https://mainnet.ckbapp.dev/rpc',
+  port: process.env.PORT || 3000,
```
Keith-CY, line 4: `3000` is set as the default port, then `fallback to 3000` at https://github.com/ckb-js/kuai/blob/develop/packages/samples/mvp-dapp/src/main.ts#L14 is redundant

Keith-CY (approved):

yanguoyu (approved):

Keith-CY:
Conflicted

Keith-CY:
The conflicts should be resolved before being merged

PR #151, opened 2023-02-15
fix: Fix request success.

yanguoyu:
It can return `inputs, outputs, outputs_data`, but need to build a tx by these data.

### DIFF
--- a/packages/samples/mvp-dapp/src/app.controller.ts
+++ b/packages/samples/mvp-dapp/src/app.controller.ts
@@ -46,7 +46,6 @@ async function getOmnilockModel(lock: Script): Promise<OmnilockModel> {
       pattern: lockHash,
       value: { type: 'register', register: { lockScript: lock, uri: actorRef.uri, pattern: lockHash } },
     })
-    return omnilockModel
   }
   return omnilockModel
 }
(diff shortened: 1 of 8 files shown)


Daryl-L (approved):

packages/samples/mvp-dapp/src/app.controller.ts
```diff
@@ -3,18 +3,22 @@ import { HexString, Script, helpers } from '@ckb-lumos/lumos'
...
 const router = new KuaiRouter()
 function createCellPattern(lock: Script) {
   return (value: UpdateStorageValue) => {
     const cellLock = value.cell.cellOutput.lock
-    return cellLock.args === lock?.args && cellLock.codeHash === lock?.codeHash && cellLock.hashType === lock?.hashType
+    return (
+      cellLock.args === lock?.args &&
+      cellLock.codeHash === lock?.codeHash &&
+      cellLock.hashType === lock?.hashType &&
+      value.cell.data === '0x'
```
Keith-CY, line 20: If the `data` is obligatory to be `0x`, the method name may be confusing because it's specific for plain omnilock cells

yanguoyu: Of cause, it's the store to save plain omnilock cells. Add filter `data==='0x'` to avoid used cells that have stored some data when calling `claim`.

Keith-CY (approved):

PR #156, opened 2023-02-20
feat: add script metadata in store

Keith-CY:
Script is the most general property of a store so it was added for facility. It can be combined with "cell pattern" later

Besides, a `meta` property which includes `capacity` is added for better user experience.

### DIFF
--- a/packages/samples/mvp-dapp/src/app.controller.ts
+++ b/packages/samples/mvp-dapp/src/app.controller.ts
@@ -35,9 +35,6 @@ function createRecordPattern(lock: Script) {
 }
 
 const getLock = (address: string) => {
-  if (!address) {
-    throw new BadRequest('invalid address')
-  }
   try {
     return helpers.parseAddress(address)
   } catch {
--- a/packages/samples/mvp-dapp/src/app.controller.ts
+++ b/packages/samples/mvp-dapp/src/app.controller.ts
@@ -34,7 +34,19 @@ function createRecordPattern(lock: Script) {
   }
 }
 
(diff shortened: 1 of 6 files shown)


packages/samples/mvp-dapp/src/app.controller.ts
```diff
@@ -34,14 +34,27 @@ function createRecordPattern(lock: Script) {
   }
 }
 
-async function getOmnilockModel(lock: Script): Promise<OmnilockModel> {
+const getLock = (address: string) => {
+  if (!address) {
+    throw new BadRequest('invalid address')
+  }
```
Daryl-L, line 40: I think this is not necessary because if the address is null, the helper.parseAddress will throw an exception.

yanguoyu (approved):

yanguoyu:
Maybe `capacity` can be designed as a `store`'s state like `data`, or `args`.

Keith-CY:
> Maybe `capacity` can be designed as a `store`'s state like `data`, or `args`.

`capacity` is probably a read-only property, not as a `state` implies which can be updated. We may add a `meta` field for these properties.

packages/samples/mvp-dapp/src/app.controller.ts
Keith-CY: the redundant condition checking was removed by https://github.com/ckb-js/kuai/pull/156/commits/5f8c5f125568ca9a77489327e9615b09f83bc218

Daryl-L (approved):

PR #157, opened 2023-02-21
feat: add script metadata in store

Keith-CY:
Script is the most general property of a store so it was added for
facility. It can be combined with "cell pattern" later

### DIFF
--- a/packages/samples/mvp-dapp/src/omnilock/omnilock.model.ts
+++ b/packages/samples/mvp-dapp/src/omnilock/omnilock.model.ts
@@ -2,7 +2,7 @@ import { CellPattern, JSONStore, OutPointString, SchemaPattern, UpdateStorageVal
 import { Cell, HexString } from '@ckb-lumos/base'
 import { BI } from '@ckb-lumos/bi'
 import { InternalServerError } from 'http-errors'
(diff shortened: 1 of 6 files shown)


Keith-CY:
This PR covers https://github.com/ckb-js/kuai/pull/157/commits/5ff58036e8951055e84c019325e26d7a8f742843, please reivew that one first

packages/samples/mvp-dapp/src/omnilock/omnilock.model.ts
```diff
@@ -64,4 +67,9 @@ export class OmnilockModel extends JSONStore<Record<string, never>> {
       witnesses: [],
     }
   }
+
+  #filterCells = () => {
+    // TODO: is this filter necessary?
```
yanguoyu, line 72: It has been filtered by `store` with cellPattern. So it's unnecessary.

Keith-CY: The filter has been removed by https://github.com/ckb-js/kuai/pull/157/commits/06e4bc2952c1a45251a5f37e562cdb1896aea97b

Daryl-L (approved):

yanguoyu (approved):

PR #158, opened 2023-02-23
fix(resource-binding): Each combination of lock script & type

Daryl-L:
script should be related to multiple store, and the filter of cell data will be added to resource binding manager later.


yanguoyu (approved):

zhengjianhui (approved):

Keith-CY:
The problem this PR tried to solve is still there.

PR #163, opened 2023-02-26
fix: fix sync on new blocks

Keith-CY:
1. fix sync on new blocks. In the previous logic, `registryOutPoint` maps to a single `model`(https://github.com/ckb-js/kuai/pull/163/files#diff-6beced91eeabe7f90f6b66349a7a50e5a2e7cac2c73c536c6c32131c8082ddcdL21), which leads to that `changes` will only be sent to a single `model` instead of all models matched by the `outPoint`, so only one `model` will be updated(https://github.com/ckb-js/kuai/pull/163/files#diff-6beced91eeabe7f90f6b66349a7a50e5a2e7cac2c73c536c6c32131c8082ddcdL147) correctly. This PR maps `registryOutPoint` to a set of `models` to make sure they are all updated by the `changes`;
2. add a `load` method in the store model, which returns the global state combined from various cells matched in the store.


PainterPuppets (approved):

packages/models/src/store/store.ts
```diff
@@ -94,7 +114,7 @@ export class Store<
   }
 
   private deserializeField(type: StorageLocation, option: unknown, value: string) {
-    this.assetStorage(this.getStorage(type))
+    this.assertStorge(this.getStorage(type))
```
yanguoyu, line 117: `Storge` means?

packages/models/src/resource-binding/manager.ts
```diff
@@ -71,112 +73,119 @@ export class Manager extends Actor<object, MessagePayload<ResourceBindingManager
...
+  ) {
+    if (cellChangeDataList.length > 0) {
+      const upgrade = [...cellChangeDataList.values()].map(([cell, witness]) => ({ cell, witness }))
+      upgrade.forEach(({ cell }) => {
+        if (!cell.outPoint) return
+        const outPointStr = outPointToOutPointString(cell.outPoint)
+        const registryMap = this.#registryOutPoint.get(outPointStr) ?? new Map()
+        registryMap.set(registry.uri, registry)
+        this.#registryOutPoint.set(outPointStr, registryMap)
+      })
```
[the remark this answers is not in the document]
Keith-CY: I would like to keep side effects outside `map` to make it clear and readable(only the result matters). The cost of iterating outputs one more time is acceptable.

packages/models/src/store/store.ts
Keith-CY: It's the method to asset `storage` is not empty. Here I just fix a typo `asset => assert` 

Ref: https://github.com/ckb-js/kuai/pull/163/files/9bc8c02c829db9dfaef273518abb4fd8b96399cc#diff-84f82286c84c4c0e853797d892f4e8a6c907a3c7298583c756b9dde3c898c8c5R325-R327

yanguoyu: Maybe `assertStorage`?

packages/models/src/store/store.ts
```diff
@@ -86,6 +88,24 @@ export class Store<
     this.#type = Reflect.getMetadata(ProviderKey.TypePattern, this.constructor)
   }
 
+  public load(path?: string) {
+    let local = {}
+    const customizer = (objValue: StructSchema, srcValue: StructSchema) => {
+      if (Array.isArray(objValue)) {
```
yanguoyu, line 94: Shall we sum values if their type is BigNumber or BI or number? BI or number is a type of molecule

packages/models/src/store/store.ts
Keith-CY: > Maybe `assertStorage`?

My fault, fixed by https://github.com/ckb-js/kuai/pull/163/commits/a536ebb88185e284fcf62bb4f80b64e61cb523ba

Daryl-L (approved):

packages/models/src/store/store.ts
Keith-CY: It's feasible, but the rule should be considered and documented explicitly, it can be a feature in the future.

Now, the duplicated field will be overridden by the latter one.

yanguoyu (approved):

PR #173, opened 2023-03-01
feat: update deploy action

PainterPuppets:


packages/samples/mvp-dapp/tsconfig.json
```diff
@@ -10,5 +10,6 @@
     "strict": true,
     "skipLibCheck": true
   },
-  "include": ["src"]
+  "files": ["src/type-extends.d.ts"],
+  "include": ["src/**/*"]
```
Keith-CY, line 14: `files` seems to be covered by `include`, do we have figured out why it doesn't work with `ts-node`?

PainterPuppets:
@Keith-CY  After checking the documentation, maybe the reason is that `ts-node` didn't read the `include`, and it works fine after adding the following code to `tsconfig.json`
```
"ts-node": {
    "files": true
  },
'```

**Comment by Keith-CY** [comment]
> @Keith-CY After checking the documentation, maybe the reason is that `ts-node` didn't read the `include`, and it works fine after adding the following code to `tsconfig.json`
> 
> ```
> "ts-node": {
>     "files": true
>   },
> ```
> 
> ref: [stackoverflow.com/a/68488480](https://stackoverflow.com/a/68488480)

I've found another solution from the StackOverflow mentioned above to enable `ts-node` without modification of code

```bash
npx ts-node --files ./src/main.ts
```

Keith-CY (approved):

yanguoyu (approved):

PR #174, opened 2023-03-01
feat(mvp): exception handle (#174

zhengjianhui:
Add a global exception handler for mvp


Daryl-L (approved):

packages/samples/mvp-dapp/tsconfig.json
```diff
@@ -10,5 +10,6 @@
     "strict": true,
     "skipLibCheck": true
   },
-  "include": ["src"]
+  "include": ["src"],
+  "files": ["src/type-extends.d.ts"]
```
yanguoyu, line 14: This can be fixed by https://github.com/ckb-js/kuai/pull/173. https://github.com/ckb-js/kuai/blob/a8579bf4bd13605f44d2548bbbd5f02f1fd04105/packages/samples/mvp-dapp/tsconfig.json#L13-L15
It should add config for ts-node

Keith-CY: This config has been added on the develop branch and could be resolved by rebase.

Ref: https://github.com/ckb-js/kuai/blob/develop/packages/samples/mvp-dapp/tsconfig.json#L14

packages/samples/mvp-dapp/src/exception.ts
```diff
@@ -0,0 +1,21 @@
+import { Context, HttpError, Next } from 'koa'
+
+export function exceptionHandle() {
+  return async (ctx: Context, next: Next) => {
+    try {
+      await next()
+    } catch (err) {
+      if (err instanceof HttpError) {
```
Keith-CY, line 8: Would it be an `Error` but not a `HttpError`?

zhengjianhui: > Would it be an `Error` but not a `HttpError`?

Users can customize business errors, so here are simply defined three cases, system errors, business errors, HttpError.

packages/samples/mvp-dapp/tsconfig.json
zhengjianhui: > This can be fixed by #173.
> 
> https://github.com/ckb-js/kuai/blob/a8579bf4bd13605f44d2548bbbd5f02f1fd04105/packages/samples/mvp-dapp/tsconfig.json#L13-L15
> 
> 
> It should add config for ts-node

Okay here I'll remove it first

packages/samples/mvp-dapp/src/exception.ts
Keith-CY: My fault, I misunderstood the nested branches.

packages/samples/mvp-dapp/src/exception.ts
```diff
@@ -0,0 +1,21 @@
...
+        ctx.body = { message: err.message }
+      } else {
+        ctx.status = 500
+        ctx.body = { message: 'Internal server error' }
+      }
+    }
+  }
+}
+
+export class BizError extends Error {}
```
Keith-CY, line 21: What if adding a biz error code in the biz error so we can make the response clear, as
HTTP code 404 => endpoint is not found
Biz code 404 => store/field is not found

zhengjianhui: > What if adding a biz error code in the biz error so we can make the response clear, as
> HTTP code 404 => endpoint is not found
> Biz code 404 => store/field is not found

That's what I was thinking, using `MvpResponse` to wrap the return value and business error, it would look something like this

```js
// definition:

export class BizError extends Error {
  errCode: string;

  constructor(message: string, errCode: string) {
    super(message);
    this.errCode = errCode;
  }
}

export class MvpResponse {
  code: string
  data: any;

  constructor(code: string, data: any) {
    this.code = code;
    this.data = data;
  }

  static ok(data: any) {
    return new MvpResponse(data, "200")
  }

  static err(data: any, code: string) {
    return new MvpResponse(data, code)
  }
}

// controller:

router.get<never, { path: string; address: string }>('/load/:address/:path', async (ctx) => {
  const { params } = ctx.payload

  if (!params?.path) {
    throw new BadRequest('invalid path')
  }

  const recordModel = await getRecordModel(params?.address)
  const value = recordModel.load(`data.${params.path}`)
  if (value) {
    ctx.ok(MvpResponse.ok(value))
  } else {
    throw new BizError('field is not found', '404')
    // or
    // ctx.ok(MvpResponse.err("field is not found", '404'))
  }
})

// err handle
export function exceptionHandle() {
  return async (ctx: Context, next: Next) => {
    try {
      await next()
    } catch (err) {
      if (err instanceof HttpError) {
        ctx.body = { message: err.message }
      } else if (err instanceof BizError) {
        ctx.status = 200
        ctx.body = MvpResponse.err(err.message, err.errCode);
      } else {
        ctx.status = 500
        ctx.body = { message: 'Internal server error' }
      }
    }
  }
}
```
Here httpCode is used for HTTP errors, 500 for the rest of the errors, and a code is used in the return value to explain the business situation

Keith-CY: > > What if adding a biz error code in the biz error so we can make the response clear, as
> > HTTP code 404 => endpoint is not found
> > Biz code 404 => store/field is not found
> 
> That's what I was thinking, using `MvpResponse` to wrap the return value and business error, it would look something like this
> 
> ```js
> // definition:
> 
> export class BizError extends Error {
>   errCode: string;
> 
>   constructor(message: string, errCode: string) {
>     super(message);
>     this.errCode = errCode;
>   }
> }
> 
> export class MvpResponse {
>   code: string
>   data: any;
> 
>   constructor(code: string, data: any) {
>     this.code = code;
>     this.data = data;
>   }
> 
>   static ok(data: any) {
>     return new MvpResponse(data, "200")
>   }
> 
>   static err(data: any, code: string) {
>     return new MvpResponse(data, code)
>   }
> }
> 
> // controller:
> 
> router.get<never, { path: string; address: string }>('/load/:address/:path', async (ctx) => {
>   const { params } = ctx.payload
> 
>   if (!params?.path) {
>     throw new BadRequest('invalid path')
>   }
> 
>   const recordModel = await getRecordModel(params?.address)
>   const value = recordModel.load(`data.${params.path}`)
>   if (value) {
>     ctx.ok(MvpResponse.ok(value))
>   } else {
>     throw new BizError('field is not found', '404')
>     // or
>     // ctx.ok(MvpResponse.err("field is not found", '404'))
>   }
> })
> 
> // err handle
> export function exceptionHandle() {
>   return async (ctx: Context, next: Next) => {
>     try {
>       await next()
>     } catch (err) {
>       if (err instanceof HttpError) {
>         ctx.body = { message: err.message }
>       } else if (err instanceof BizError) {
>         ctx.status = 200
>         ctx.body = MvpResponse.err(err.message, err.errCode);
>       } else {
>         ctx.status = 500
>         ctx.body = { message: 'Internal server error' }
>       }
>     }
>   }
> }
> ```
> 
> Here httpCode is used for HTTP errors, 500 for the rest of the errors, and a code is used in the return value to explain the business situation

Upvote

packages/samples/mvp-dapp/src/response.ts
```diff
@@ -0,0 +1,17 @@
+export class MvpResponse {
+  code: string
+  data: unknown
```
Keith-CY, line 3: Declaring the data as a generic type would be more user friendly

packages/samples/mvp-dapp/src/exception.ts
```diff
@@ -0,0 +1,29 @@
+import { Context, HttpError, Next } from 'koa'
+import { MvpResponse } from './response'
+
+export function exceptionHandle() {
```
Keith-CY, line 4: `exceptionHandler` or `handleException`

packages/samples/mvp-dapp/src/exception.ts
```diff
@@ -0,0 +1,29 @@
...
+      } else {
+        ctx.status = 500
+        ctx.body = { message: 'Internal server error' }
+      }
+    }
+  }
+}
+
+export class BizError extends Error {
+  errCode: string
```
Keith-CY, line 23: `code` was used in `MvpResponse`, would it be better to name them in a consistent style?

packages/samples/mvp-dapp/src/response.ts
zhengjianhui: Okay, I'll refactor it

> Declaring the data as a generic type would be more user friendly

Okay, I'll refactor it

packages/samples/mvp-dapp/src/exception.ts
zhengjianhui: Okay, I'll refactor it

> `exceptionHandler` or `handleException`

Okay, I'll refactor it

packages/samples/mvp-dapp/src/exception.ts
zhengjianhui: > `code` was used in `MvpResponse`, would it be better to name them in a consistent style?

Okay, I'll refactor it

Keith-CY: The point I referred is the name style, `code` used in `MvpResponse` while `errCode` is used in `MvpError`

zhengjianhui: > The point I referred to is the name style, `code` used in `MvpResponse` while `errCode` is used in `MvpError`

Oh, I see, the actual `code` and `errCode` are similar

packages/samples/mvp-dapp/src/exception.ts
```diff
@@ -0,0 +1,29 @@
...
+      await next()
+    } catch (err) {
+      if (err instanceof HttpError) {
+        ctx.body = { message: err.message }
+      } else if (err instanceof MvpError) {
+        ctx.status = 200
+        ctx.body = MvpResponse.err(err.message, err.code)
+      } else {
+        ctx.status = 500
+        ctx.body = { message: 'Internal server error' }
```
yanguoyu, line 16: Shall we need to print the error or return the detailed error when in the development stage?
It may easy to locate some unexpected exceptions.

Keith-CY: It's a MVP DApp, we may leave it simple

yanguoyu (approved):

Keith-CY:
Commits should be signed as https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

PR #190, opened 2023-03-15
feat: add builtin test action

PainterPuppets:
- add builtin test action
 - add simple test for kuai-mvp-dapp

### DIFF
--- a/packages/core/jest.config.js
+++ b/packages/core/jest.config.js
@@ -2,5 +2,5 @@
 module.exports = {
   preset: 'ts-jest',
   testEnvironment: 'node',
-  testPathIgnorePatterns: ['/__fixtures__/', '/__utils__/'],
-};
+  testPathIgnorePatterns: ['/__fixtures__/', '/__utils__/', 'lib/builtin-tasks/test.js', 'src/builtin-tasks/test.ts'],
+}
--- a/packages/core/src/builtin-tasks/test.ts
+++ b/packages/core/src/builtin-tasks/test.ts
@@ -1,0 +1,52 @@
+import { runCLI } from '@jest/core'
+import type { Config as JestConfig } from 'jest'
+import path from 'node:path'
+import type { AggregatedResult } from '@jest/test-result'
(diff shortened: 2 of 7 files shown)


packages/core/jest.config.js
```diff
@@ -2,5 +2,5 @@
 module.exports = {
   preset: 'ts-jest',
   testEnvironment: 'node',
-  testPathIgnorePatterns: ['/__fixtures__/', '/__utils__/'],
```
Keith-CY, line 5: why is `src` ignored

PainterPuppets: because the `test.js` file is included by `jest` by default, while the `src` include test file may be  unexpected.

Keith-CY: Sure that `src` should not include any test files, but `ignore src` doesn't restrict the improper behavior.

packages/core/src/builtin-tasks/test.ts
yanguoyu, line 1: This task is used to run the project's test cases?

Keith-CY: > This task is used to run the project's test cases?

It's expected to trigger tests in projects created by `kuai cli`

yanguoyu: What is the difference between running tests by `yarn test` directly?

Keith-CY: We’re going to standardize workflow among projects, and provide preset scripts for daily actions, tests can be in multiple modes(for backend, for rust, in watch mode, generate coverage, etc.), so we’d better deliver them from the beginning

Besides, configuring tests is repeated and boring, provided by `cli` would be nice to developers, especially those who are not familiar with JavaScript ecosystem.

yanguoyu (approved):

packages/core/jest.config.js
Keith-CY: Any update on the `ignore specific test.ts file in src directory`?

Daryl-L (approved):

PR #191, opened 2023-03-15
feat: add controller for io

PainterPuppets:
- add `Get` / `Post` / `Delete` / `Put` decorator 
 - add `Header` / `Param` / `Body` / `Query` decorator
 - add prefix path for controller
 - add controller class


packages/io/__tests__/controller.ts
yanguoyu, line 1: How about add a prefix on the Class?
Ex.
```typescript
@Controller('app')
class AppController extends BaseController {
  @Get()
  async hello(ctx: Context) {
    ctx.ok('hello root')
  }
  @Get('/parent')
  async parent(ctx: Context) {
    ctx.ok('hello parent')
  }
}
```
Then the `AppController.hello`'s path will be `/app`, and the `AppController.parent`'s path will be `/app/parent`

packages/io/src/controller.ts
```diff
@@ -0,0 +1,87 @@
...
+  public middleware(): Middleware<RouterContext> {
+    return async (ctx, next) => {
+      const payload = ctx.payload
+
+      if (!isRoutePayload(payload)) {
+        return next()
+      }
+
+      const route = this.getRoutes().find((route) => route.method === payload.method && matchPath(payload.path, route))
+      if (route) {
```
yanguoyu, line 73: Maybe `route` can cache with `payload.method` and `payload.path`?

packages/io/__tests__/controller.ts
```diff
@@ -0,0 +1,52 @@
...
+    ctx.ok('hello parent')
+  }
+
+  @Get('/parent/children')
+  async children(ctx: Context) {
+    ctx.ok('hello children')
+  }
+
+  @Get('/:username')
+  async username(ctx: Context<{ payload: { params: { username?: string } } }>) {
```
Daryl-L, line 23: How about using param decorators to inject the parameters.

```
class AppController {
  @Get('/:username')
  async username(@Query('age') age: number, @Param('username') username: string) {}
}
```

PainterPuppets: 👌

packages/io/src/decorator.ts
```diff
@@ -0,0 +1,31 @@
+import { KUAI_ROUTE_METADATA_METHOD, KUAI_ROUTE_METADATA_PATH } from './metadata'
+import 'reflect-metadata'
+
+export function Get(path = '/'): MethodDecorator {
```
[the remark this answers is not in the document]
PainterPuppets: `MethodDecorator` is typescript built-in type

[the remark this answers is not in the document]
Keith-CY: Got it, I was confused because `get`, `post` are also `methods` in other aspects

packages/io/src/controller.ts
```diff
@@ -45,7 +45,69 @@ const getMethods = (obj: object): string[] => {
...
+              if (ctx.payload.query === undefined) {
+                return undefined
+              }
+              if (arg.data === undefined) {
+                return ctx.payload.query
+              }
+              return ctx.payload.query[arg.data]
+            }
+
+            return undefined
```
yanguoyu, line 94: I think this code snippet can be cleaner. 
```typescript
let typeValue: any
switch (arg.paramtype):
    case RouteParamtypes.BODY:
         typeValue = ctx.payload.body
         break;
    //other cases
if (arg.data !== undefined) return typeValue?.[arg.data]
return typeValue
```

packages/io/src/decorator.ts
```diff
@@ -25,6 +30,52 @@ export function Delete(path = '/'): MethodDecorator {
...
 
+export type ParamData = string | number
+export interface RouteParamMetadata {
+  index: number
+  paramtype: RouteParamtypes
+  data?: ParamData
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function assignMetadata<TArgs = any>(args: TArgs, paramtype: RouteParamtypes, index: number, data?: ParamData) {
```
yanguoyu, line 27: Why `TArgs` type need to set default with `any`?

packages/io/src/helper.ts
```diff
@@ -0,0 +1,7 @@
+export const addLeadingSlash = (path?: string): string =>
+  path && typeof path === 'string' ? (path.charAt(0) !== '/' ? '/' + path : path) : ''
+
+export const concatPaths = (a: string, b: string) =>
+  addLeadingSlash(stripEndSlash(a)) + addLeadingSlash(stripEndSlash(b))
+
+export const stripEndSlash = (path: string) => (path[path.length - 1] === '/' ? path.slice(0, path.length - 1) : path)
```
Keith-CY, line 7: These methods should be tested

packages/io/src/decorator.ts
```diff
@@ -0,0 +1,105 @@
...
+    if (typeof prefixOrOptions === 'string') {
+      return prefixOrOptions
+    }
+
+    if (typeof prefixOrOptions === 'object') {
+      return prefixOrOptions.prefix || '/'
+    }
+
+    return '/'
+  })()
```
Keith-CY, line 66: Will 
```js
const path = typeof prefixOrOptions === 'string' ? prefixOrOptions : prefixOrOptions?.prefix || '/'
``` 
work the same?

packages/io/src/controller.ts
```diff
@@ -0,0 +1,156 @@
...
+  let currentObj = obj
+  do {
+    Object.getOwnPropertyNames(currentObj).map((item) => properties.add(item))
+  } while ((currentObj = Object.getPrototypeOf(currentObj)))
+  return [...properties.keys()].filter((item) => typeof obj[item as keyof typeof obj] === 'function')
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export class BaseController {
+  _routes: Route[] = this.getRoutes()
```
Keith-CY, line 52: Is `_routes` intended to be a private field? If so, `#routes` could be used

yanguoyu (approved):

Keith-CY (approved):

Issue #101, opened 2023-01-09
Integrate Nexus into the mvp demo


Keith-CY:
The MVP DApp is expected to be [work with MetaMask](https://github.com/ckb-js/kuai/issues/54#issuecomment-1365630586) at this early stage.

Meanwhile, a browser extension wallet of Nervos CKB, named Nexus(https://github.com/ckb-js/nexus) is under development.

Once Nexus is ready for tests, we will try to support it in the MVP DApp.

Keith-CY:
We're considering integrating Nexus into the Kuai MVP DApp(https://github.com/ckb-js/kuai/issues/54), any documents would you recommend for us to start? @homura 

Online demo: https://kuai-mvp-dapp-ui.vercel.app/

homura:
Nexus don't currently have a friendly document, only a [pre-design](https://github.com/ckb-js/nexus/blob/pre-design/docs/design.md#javascript-provider) in the PR, we plan to provide a [demo dApp](https://github.com/zhb666/Nervos-DAO) of Nervos DAO to show how to use Nexus. Nexus MVP will probably be released around mid-March

Keith-CY:
> Nexus don't currently have a friendly document, only a [pre-design](https://github.com/ckb-js/nexus/blob/pre-design/docs/design.md?rgh-link-date=2023-02-01T06%3A56%3A36Z#javascript-provider) in the PR, we plan to provide a [demo dApp](https://github.com/zhb666/Nervos-DAO) of Nervos DAO to show how to use Nexus. Nexus MVP will probably be released around mid-March

Cool, looking forward to it

Keith-CY:
Tmp ddl: 03/08

Keith-CY:
Demo repo: https://github.com/zhangyouxin/demo-nexus/

Keith-CY:
Connect to nexus: https://github.com/zhangyouxin/demo-nexus/blob/5b25da92df19487511b0e87a293ec660579d169f/pages/index.tsx#L231-L236
Create next receiving address: https://github.com/zhangyouxin/demo-nexus/blob/5b25da92df19487511b0e87a293ec660579d169f/pages/index.tsx#L243-L250
Sign transaction of full ownership lock: https://github.com/zhangyouxin/demo-nexus/blob/5b25da92df19487511b0e87a293ec660579d169f/pages/index.tsx#L180-L183

Keith-CY:
There're 3 obstacles on integrating nexus

[...]

PR #220, opened 2023-04-08
wip: integrate with nexus

Keith-CY:
Should work with https://github.com/Magickbase/kuai-mvp-dapp-ui/pull/13, but not verified because https://github.com/ckb-js/kuai/issues/101#issuecomment-1499900320


yanguoyu:
I only find some delete codes. Do you lose some codes?

Keith-CY:
> I only find some delete codes. Do you lose some codes?

Modification on the `view` is enough, in fact. The model name `omnilock model` is left intact because there's another update(https://github.com/ckb-js/kuai/pull/181) on it, and this PR will be rebased on that one.

image

yanguoyu (approved):

PR #223, opened 2023-04-10
fix(cor): global exception handle

zhengjianhui:
Fix the default global exception handling


yanguoyu (approved):

Keith-CY:
Please add the reason for removing the default global exception handler

Keith-CY (approved):

zhengjianhui:
We can merge first, which can fix the bug in the current MVP. I will submit a PR later to supplement the logic of cor.useExceptionHandler.

Keith-CY:
> We can merge first, which can fix the bug in the current MVP. I will submit a PR later to supplement the logic of `cor.useExceptionHandler`.

Sure, there're already 2 approvals, and this PR can be merged freely.

But it's blocked by `unsigned commits`, please have a check and force push this one

Keith-CY:
These commits are still unsigned, which blocks the merge.
Please follow this guide(https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) to sign commits

PR #224, opened 2023-04-10
feat: Merged state and user can change merged state.

yanguoyu:
1. Extract merge strategy to `merge-strategy`, and create `DefaultMergeStrategy` and `UseLatestStrategy` .
2. And now `mergeState` has calculated when `handleCell`, and developers can change `mergeState` by `set`.


packages/models/src/store/merge-strategy.ts
```diff
@@ -0,0 +1,266 @@
...
+      return {
+        update: [
+          {
+            outPointString: lastOutPoint,
+            state: updateInfo.value as S,
+          },
+        ],
+      }
+    }
+    const updateValueInstates = get([lastOutPoint, ...updateInfo.paths.slice(0, updateInfo.paths.length - 1)])(
```
Keith-CY, line 53: `updateValueInstates` can be `updateValueInStates`

packages/models/src/store/merge-strategy.ts
```diff
@@ -0,0 +1,266 @@
...
+      updateInfo.states,
+    )
+    if (updateValueInstates === undefined) {
+      throw new NonExistentException(`${lastOutPoint}${updateInfo.paths.join('.')}`)
```
Keith-CY, line 57: A separator is desirable between `out point` and `update info`

packages/models/src/store/merge-strategy.ts
```diff
@@ -0,0 +1,266 @@
...
+  private defaultMergeWith = mergeWith((objValue, srcValue) => {
+    if (Array.isArray(objValue)) {
+      return objValue.concat(srcValue)
+    }
+  })
+
+  merge(object: S, source?: S): S {
+    if (source === undefined || source === null) return object
+    if (Array.isArray(object)) return object.concat(source) as S
+    return this.defaultMergeWith(source, object)
```
Keith-CY, line 82: The order of `destination` and `source` is reversed from that of definition

packages/models/src/store/merge-strategy.ts
```diff
@@ -0,0 +1,266 @@
...
+  ): {
+    removePath?: string[]
+    state?: S
+  } {
+    if (typeof state !== 'object' || state === null || !paths.length)
+      return { state: value as S, removePath: Array.isArray(value) ? paths : undefined }
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    let current: any = state
+    for (let idx = 0; idx < paths.length; idx++) {
+      if (paths[idx] in current) {
```
Keith-CY, line 98: use `continue` to avoid nested `if` statements

packages/models/src/store/merge-strategy.ts
```diff
@@ -0,0 +1,266 @@
...
+          const lastLevelPath = currentPaths.slice(0, -1)
+          const lastLevelPathvalue = get(lastLevelPath)(state)
+          // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+          delete lastLevelPathvalue[currentPaths.at(-1)!]
+        } else {
+          // ignore if root is simple type, no need to handle
+        }
+        return false
+      }
+      return true
```
Keith-CY, line 158: The `continueDeepFor` should be defined independently for testing and understanding

packages/models/src/store/merge-strategy.ts
```diff
@@ -0,0 +1,266 @@
...
+    value: unknown
+    outPointStrings: OutPointString[]
+    states: Record<OutPointString, S>
+    isValueEqual(value: unknown, compare: unknown): boolean
+    isSimpleType(value: unknown): boolean
+  }): {
+    update?: {
+      outPointString: OutPointString
+      state: S
+    }[]
```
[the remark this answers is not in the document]
yanguoyu: It's better to use `Array` because we'd better keep the order for `outPointString `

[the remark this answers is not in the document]
Keith-CY: Gotcha

Keith-CY:
Please have a review @zhengjianhui @Daryl-L @PainterPuppets

packages/models/src/store/merge-strategy.ts
yanguoyu: It depends on the outer variable `mergedState`. So I will add test cases for the `removeNotMergedState` function.

packages/models/src/store/merge-strategy.ts
yanguoyu: We should use the new value in `object` to cover the old value in `source`, so we need set `object` as the second parameter.

Keith-CY: If we can switch the parameters of `private defaultMergeWith` so they won't be mixed in all cases.

packages/models/src/store/merge-strategy.ts
Keith-CY: If we can pass `outer variables` into it as a parameter to make the logic independent and easy to test

Keith-CY (approved):

Keith-CY:
Please have a review @zhengjianhui @Daryl-L @PainterPuppets

Daryl-L (approved):

PainterPuppets (approved):

Keith-CY:
Please resolve the conflicts

yanguoyu:
I'm adding test cases for `merge-strategy`.

PR #242, opened 2023-04-19
feat: add deploy contract command

PainterPuppets:
-  add deploy contract command

image

### deploy contract workflow:
 - 1. `kuai contract new --name [contract-name]`
image

 - 2. `kuai contract build --release`
 
image

 - 3. `kuai contract deploy --name [contract-name] --from [0x.....] --signer ckb-cli`
image

#### Tips: How to deploy contract to other network
 - `kuai contract deploy --name [contract-name] --from [0x.....] --network [networkname]` 
 -  or set network to `.env` file

#### Tips: add more network config 
Kuai currently has three default settings for networks,` testnet` / `mainnet` / `devnet`, and users can also add custom network 
 - edit `kuai.config.js` of project and add network tor `networks` field
``` typescript
networks: {
    [networkname: string]: {
        rpcUrl: string
        prefix: string
        scripts?: Record<string, ScriptConfig>
    }
}
```


packages/core/src/builtin-tasks/contract.ts
```diff
@@ -0,0 +1,209 @@
...
+    const res = execSync('ckb-cli account list --output-format json').toString()
+    const accountList: Array<{ lock_arg: string }> = JSON.parse(res)
+
+    if (accountList.some((account) => account.lock_arg === lockArg)) {
+      return
+    }
+
+    fs.writeFileSync('./.tmp_pk', privateKey)
+
+    console.log('add the user to ckb-cli, enter a password for the signature please')
```
Keith-CY, line 52:
```suggestion
    console.info('add the user to ckb-cli, enter a password for the encryption please')
```

packages/core/src/builtin-tasks/contract.ts
```diff
@@ -0,0 +1,209 @@
...
+    execSync('ckb-cli account import --privkey-path ./.tmp_pk', { stdio: 'inherit' })
+
+    fs.rmSync('./.tmp_pk')
```
Keith-CY, line 55: It should be wrapped by `try cache` and remove the temp file in the `finally` block to make sure it will be erased

Keith-CY: Besides, it's better to use a random file name to avoid predictable file reading.

packages/core/src/builtin-tasks/contract.ts
```diff
@@ -0,0 +1,209 @@
...
+  fee: number
+  env: string
+  migrate: string
+}
+
+subtask('contract:deploy')
+  .addParam('name', 'name of the contract to be deployed', '', paramTypes.string, true)
+  .addParam(
+    'chain',
+    'ckb rpc url or name [default: dev]  [possible values: dev, test, mainnet, http://localhost:8114 etc...]',
```
Keith-CY, line 106: Better to name them clearly. `test` is for `testnet` but may be misleading for others to regard it as a local test environment

packages/core/src/builtin-tasks/contract.ts
```diff
@@ -0,0 +1,209 @@
...
+    'dev',
+    paramTypes.string,
+    true,
+  )
+  .addParam('isMainnet', '', false, paramTypes.boolean, true)
+  .addParam(
+    'fee',
+    "Per transaction's fee, deployment may involve more than one transaction. default: [1]",
```
Keith-CY, line 114:
```suggestion
    "Per transaction's fee, deployment may involve more than one transaction. default: [1] shannon/Byte",
```

packages/core/src/builtin-tasks/contract.ts
```diff
@@ -0,0 +1,209 @@
...
+  )
+  .addParam(
+    'migrate',
+    'Use previously deployed cells as inputs. [default: on]  [possible values: on, off]',
+    'on',
+    paramTypes.string,
+    true,
+  )
+  .setAction(async ({ name, chain, fee, env, migrate, ...args }: DeployArgs, { config, run }) => {
+    const isMainnet = chain === 'mainnet' ? true : args.isMainnet
```
Keith-CY, line 134: How about only using one field to determine `isMainnet`, otherwise the priority should be clear

packages/core/src/builtin-tasks/contract.ts
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    if (!deployerPrivateKey) {
+      throw new Error('please set deployerPrivateKey in kuai.config.js')
+    }
+
+    await run('contract:import-private-key', { privateKey: deployerPrivateKey })
+
+    const deployerLock: Script = {
+      codeHash: '0x9bd7e06f3ecf4be0f2fcd2188b23f1b9fcc88e5d4b65a8637b17723bbda3cce8',
```
Keith-CY, line 150: Magic string, could be assigned to a constant first

packages/core/src/builtin-tasks/contract.ts
```diff
@@ -0,0 +1,209 @@
...
+
+    const generateContractDeployConfig = (name: string, enableTypeId = true) => {
+      return {
+        name,
+        enable_type_id: enableTypeId,
+        location: { file: `build/release/${name}` },
+      }
+    }
+
+    deploymentConfig['cells'] = [generateContractDeployConfig(deployedContractName)]
```
Keith-CY, line 174: Unnecessary function call

packages/core/src/builtin-tasks/contract.ts
```diff
@@ -0,0 +1,209 @@
...
+
+    fs.writeFileSync(path.join(workspace, 'deployment.toml'), TOML.stringify(deploymentConfig))
+
+    const ckbRpcUrl = (() => {
+      if (chain === 'dev') {
+        return 'http://localhost:8114'
+      }
+
+      if (chain === 'test') {
+        return 'https://testnet.ckb.dev/rpc'
```
Keith-CY, line 189: Better to load these `urls` from configuration

packages/core/src/type/runtime.ts
```diff
@@ -12,13 +12,27 @@ export interface KuaiConfig {
...
+
+export type ContractConfig = {
+  workspace?: string
+  deployment?: {
+    type?: string
+    deployerPrivateKey: string
+    lock?: {
+      codeHash: string
+      args: string
+      hashType: string
```
yanguoyu, line 26:
```suggestion
      hashType: 'data' | 'type' | 'data1'
```

packages/core/src/type/runtime.ts
```diff
@@ -12,13 +12,27 @@ export interface KuaiConfig {
...
   jest?: JestConfig
+  contract?: ContractConfig
+}
+
+export type ContractConfig = {
+  workspace?: string
+  deployment?: {
+    type?: string
+    deployerPrivateKey: string
+    lock?: {
```
Daryl-L, line 23: Maybe use `Script` defined in `lumos` is much better?

packages/core/src/builtin-tasks/contract.ts
```diff
@@ -0,0 +1,209 @@
...
+    true,
+  )
+  .addParam(
+    'migrate',
+    'Use previously deployed cells as inputs. [default: on]  [possible values: on, off]',
+    'on',
+    paramTypes.string,
+    true,
+  )
+  .setAction(async ({ name, chain, fee, env, migrate, ...args }: DeployArgs, { config, run }) => {
```
Daryl-L, line 133: How about extracting the code of deployment so that we can use it to deploy the infra scripts?

Keith-CY:
How is it going

PainterPuppets:
The contract module of kuai needs a way to manage and organize, and some parts of the current contract management by `capsule` do not meet the needs

For example:
	- `capsule` has its own directory structures, which can cause some confusion when using `kuai`
	- deploy contracts by `capsule` can only be signed by `ckb-cli`, which can be a strange process
	- It is not possible to deploy multiple contracts in a single transaction

So `kuai` needs to deploy contract code through `lumos` in order to meet the flexible deploy methods, which will make `capsule`'s ability to organize contracts invalid, so `kuai` itself needs a way to organize contracts

For this reason, the following three structures are defined, where the `ContractManager` is used to manage contract information and some configuration. The `Contract` is the entity for individual contract information, and `KuaiContractMigration` is generated after each deployment to track the deployment


```ts
export interface ContractManager {
	contracts: KuaiContract[]

	new: (name: string, template?: 'rust' | 'c' | 'c-sharedlib')
	build: (name: string) => void
	createDeployTx: (name: string, deployer: Address) => Promise<Transaction>
	createUpgradeTx: (name: string, deployer: Address) => Promise<Transaction>
	commitTx: (name: string, witnesses: string[]) => Promise<Hash>
}

interface KuaiContract {
	name: string
	templateType: string
	srcPath: string
	binPath: string
	deployType: 'data' | 'typeId'
	migrations: KuaiContractMigration[]
}

interface KuaiContractMigration {
	chain: string
	txHash: Hash
	index: number
	dataHash: Hash
	typeId: Hash
}
```


Specific implementation details:
When call `new`, use `capsule` to generate the code by template
When call `build`, use `capsule` to build the contract file (perhaps in the future can support direct build)

deploy process:
`kuai-cli` calls `createDeployTx` of the `ContractManager` to generate the tx, which is then signed by a user-selected signature method (e.g. ckb-cli / metamask / other...)
`kuai-cli` calls the `ContractManager`'s `commitTx` to commit the tx to the chain

yanguoyu:
> kuai-cli calls createDeployTx of the ContractManager to generate the tx, which is then signed by a user-selected signature method (e.g. ckb-cli / metamask / other...)

I think we can support signed by `kuai`, then developers will not need to know more tools.

Keith-CY:
> > kuai-cli calls createDeployTx of the ContractManager to generate the tx, which is then signed by a user-selected signature method (e.g. ckb-cli / metamask / other...)
> 
> I think we can support signed by `kuai`, then developers will not need to know more tools.

Upvote, only `kuai-cli` should be exposed to reduce confusion.

PainterPuppets:
> > kuai-cli calls createDeployTx of the ContractManager to generate the tx, which is then signed by a user-selected signature method (e.g. ckb-cli / metamask / other...)
> 
> I think we can support signed by `kuai`, then developers will not need to know more tools.

Does support for signature by "kuai" mean wrapping "ckb-cli" or using the keystore directly?

Keith-CY:
> > > kuai-cli calls createDeployTx of the ContractManager to generate the tx, which is then signed by a user-selected signature method (e.g. ckb-cli / metamask / other...)
> > 
> > 
> > I think we can support signed by `kuai`, then developers will not need to know more tools.
> 
> Does support for signature by "kuai" mean wrapping "ckb-cli" or using the keystore directly?

It's about the detail of implementation, but `ckb-cli` requires private key on the disk, which is not recommended.

Keith-CY:
How is it going @PainterPuppets

PainterPuppets:
Have implemented the ContractManager part, is doing the deployment of the signature part, because the keystore built into the kuai spend more time, so are packing ckb-cli signature part, is expected to be completed and submit tomorrow

@Keith-CY

Keith-CY:
How is it going

packages/core/src/builtin-tasks/contract.ts
PainterPuppets: https://github.com/ckb-js/kuai/pull/242/files#diff-5cc22f628d3f35959679ab626647dec8f2027fad3513e367efb42735d69c50cbR66-R71

Keith-CY:
Please update the usage in the PR message and I'll request some reviews from outside teams that may use `kuai` to deploy contracts.

PainterPuppets:
> Please update the usage in the PR message and I'll request some reviews from outside teams that may use `kuai` to deploy contracts.

Updated, only the ckb-cli signature method is supported for now, but the extension point is left, I don't know what the outside teams is using for the signature, I can go add it

Keith-CY:
> > Please update the usage in the PR message and I'll request some reviews from outside teams that may use `kuai` to deploy contracts.
> 
> Updated, only the ckb-cli signature method is supported for now, but the extension point is left, I don't know what the outside teams is using for the signature, I can go add it

Should kuai's user keep their private key on the disk to sign the transaction with `ckb-cli`?

Keith-CY:
> > > Please update the usage in the PR message and I'll request some reviews from outside teams that may use `kuai` to deploy contracts.
> > 
> > 
> > Updated, only the ckb-cli signature method is supported for now, but the extension point is left, I don't know what the outside teams is using for the signature, I can go add it
> 
> Should kuai's user keep their private key on the disk to sign the transaction with `ckb-cli`?
> 
> If so, signing a transaction with lumos should be considered, and the private key should be decrypted from a keystore file instead of reading a private key file

But it could be treated as a low-priority one, signing a transaction with `secp256k1_blake160_multisig_all` is more desired

PainterPuppets:
> > > Please update the usage in the PR message and I'll request some reviews from outside teams that may use `kuai` to deploy contracts.
> > 
> > 
> > Updated, only the ckb-cli signature method is supported for now, but the extension point is left, I don't know what the outside teams is using for the signature, I can go add it
> 
> Should kuai's user keep their private key on the disk to sign the transaction with `ckb-cli`?

No, if sign with `ckb-cli`, need to import account in `ckb-cli`, but this part is the logic of `ckb-cli`,  after that it should be possible to extend more signature methods, like `metamask` / `Ledger` / other...

Keith-CY:
> > > > Please update the usage in the PR message and I'll request some reviews from outside teams that may use `kuai` to deploy contracts.
> > > 
> > > 
> > > Updated, only the ckb-cli signature method is supported for now, but the extension point is left, I don't know what the outside teams is using for the signature, I can go add it
> > 
> > 
> > Should kuai's user keep their private key on the disk to sign the transaction with `ckb-cli`?
> 
> No, if sign with `ckb-cli`, need to import account in `ckb-cli`, but this part is the logic of `ckb-cli`, after that it should be possible to extend more signature methods, like `metamask` / `Ledger` / other...

I see, so `kuai` doesn't care about where the private key in ckb-cli is from, it's reasonable

PainterPuppets:
> If so, signing a transaction with lumos should be considered, and the private key should be decrypted from a keystore file instead of reading a private key file

I thought ahead about the multisg implementation, the specific logic part does not conflict with multisig, at the cli level it will probably look like this

`kuai contract deploy --name [contract-name] --address [multisig_1] --address [multisig_2] --address [multisig_3] --network [networkname]`

packages/core/package.json
```diff
@@ -10,6 +10,11 @@
   "dependencies": {
     "@ckb-js/kuai-common": "^0.0.1",
     "@ckb-js/kuai-docker-node": "0.0.1",
+    "@ckb-lumos/base": "0.19.0",
+    "@ckb-lumos/config-manager": "0.19.0",
+    "@ckb-lumos/hd": "0.19.0",
+    "@ckb-lumos/helpers": "0.20.0-alpha.2",
```
yanguoyu, line 16: Why not use a unified version?

packages/core/src/ckbcli.ts
yanguoyu, line 1: Should we rename file to `ckb-cli`?

packages/core/src/util/chain.ts
yanguoyu, line 1: The tx information in block 0 is stable, can we use const variables to define this information?

packages/core/src/constants.ts
```diff
@@ -1,5 +1,29 @@
...
+    prefix: 'ckb',
+    scripts: config.predefined.LINA.SCRIPTS,
+  },
+  testnet: {
+    rpcUrl: 'https://testnet.ckb.dev/rpc',
+    prefix: 'ckt',
+    scripts: config.predefined.AGGRON4.SCRIPTS,
+  },
+  devnet: {
+    rpcUrl: 'http://localhost:8114',
```
Keith-CY, line 22: Better to use `127.0.0.1` because `localhost` will be resolved to `ipv6` since Node@17

Ref: https://github.com/nodejs/node/pull/39987

packages/core/src/errors-list.ts
```diff
@@ -42,11 +42,23 @@ Please run: npm install --save-dev ts-node`,
...
       code: 'UNSUPPORTED_NETWORK',
       message: 'Unsupported network %var%.',
     },
+    UNSUPPORTED_SIGNER: {
+      code: 'UNSUPPORTED_SIGNER',
+      message: 'Unsupported signer %var%.',
+    },
+    CONTRACT_RELEASE_FILE_NOT_FOUND: {
+      code: 'CONTRACT_RELEASE_FILE_NOT_FOUND',
+      message: '%var% contract release file not found, please check conrtact is exists or build it first',
```
Keith-CY, line 61:
```suggestion
      message: '%var% contract release file not found, please check if conrtact exists or build it first',
```

packages/core/src/contract.ts
```diff
@@ -0,0 +1,67 @@
...
+type Signature = string
+export type MessageSigner = (message: string, fromInfo: FromInfo) => Promise<Signature>
+
+export class ContractDeployer {
+  #rpc: RPC
+  #index: Indexer
+  #config: config.Config
+
+  constructor(
+    private readonly signer: MessageSigner,
```
Keith-CY, line 17: How about using private field instead of private declaration

packages/core/src/util/chain.ts
```diff
@@ -0,0 +1,49 @@
+import { RPC, config, utils } from '@ckb-lumos/lumos'
+
+// Special cells in genesis transactions: (transaction-index, output-index)
+const SpecialCellLocation: {
```
Keith-CY, line 4: Better to add a reference of this locations

packages/core/src/util/chain.ts
```diff
@@ -0,0 +1,49 @@
...
+      HASH_TYPE: 'type',
+      ...generateGenesisScriptConfig(...SpecialCellLocation['DAO']),
+      DEP_TYPE: 'code',
+    },
+    SECP256K1_BLAKE160: {
+      HASH_TYPE: 'type',
+      ...generateGenesisScriptConfig(...SpecialCellLocation['SIGHASH_GROUP']),
+      CODE_HASH: generateGenesisScriptConfig(...SpecialCellLocation['SIGHASH']).CODE_HASH,
+      DEP_TYPE: 'depGroup',
+      SHORT_ID: 0,
```
Keith-CY, line 39: What does `SHORT_ID` mean here?

packages/core/src/util/chain.ts
PainterPuppets: I'm not quite sure where this reference can be found, these magic location is what I see in the source code of ckb-cli

packages/core/src/util/chain.ts
PainterPuppets: SHORT_ID is required for lumos to parse short address

Keith-CY: > SHORT_ID is required for lumos to parse short address

I think it can be removed if it's not mandatory in lumos because short address is deprecated, and we should guide users not to use it.

packages/core/src/util/chain.ts
Keith-CY: Please refer to the code of ckb-cil, and I'll open an issue in ckb-cli repo for detailed info

PainterPuppets: https://github.com/nervosnetwork/ckb-sdk-rust/blob/0d6cbb38f59d6bf21982157874c17a1e71b83c81/src/constants.rs#L20-L24

Keith-CY: My bad, I didn't notice the comment above, which clearly implies the definition of `[number, number]` is `[transaction-index, output-index]`

Among them, `SIGNHASH_GROUP`, `MULTISIG_GROUP`, `DAO` can be found at https://github.com/nervosnetwork/rfcs/blob/master/rfcs/0024-ckb-genesis-script-list/0024-ckb-genesis-script-list.md
`SIGHASH`, `MULTISIG` are not used normally. I'm not sure if these two will be used in our project.

Besides, we can use labeled tuple for readability

Ref: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-0.html#labeled-tuple-elements

yanguoyu (approved):

yanguoyu (approved):

Keith-CY:
I see [feat: kuai contract support multisig deploy](https://github.com/ckb-js/kuai/pull/242/commits/da10bc89a65b26c630ee0c133640d342839efc7d) has been merged, and an example of 0/1 multisig tx was demonstrated in that PR. 

My question is how to share the partially signed transaction via `kuai`, e.g. the contract should be deployed by 2/3 multisig, how should the 2 users to deploy the contract.

And the example could be appended in this PR because the PR of [feat: kuai contract support multisig deploy](https://github.com/ckb-js/kuai/pull/242/commits/da10bc89a65b26c630ee0c133640d342839efc7d) will be inconspicuous when it's one commit of this PR

Keith-CY:
> I see [feat: kuai contract support multisig deploy](https://github.com/ckb-js/kuai/pull/242/commits/da10bc89a65b26c630ee0c133640d342839efc7d) has been merged, and an example of 0/1 multisig tx was demonstrated in that PR.
> 
> My question is how to share the partially signed transaction via `kuai`, e.g. the contract should be deployed by 2/3 multisig, how should the 2 users to deploy the contract.
> 
> And the example could be appended in this PR because the PR of [feat: kuai contract support multisig deploy](https://github.com/ckb-js/kuai/pull/242/commits/da10bc89a65b26c630ee0c133640d342839efc7d) will be inconspicuous when it's one commit of this PR

This feature is required by other features(deploy contract on devnet, complete workflow of building a dapp) so let's merge it first, other details could be added later

Issue #235, opened 2023-04-12
Support infrastructural scripts on devnet


Keith-CY:
The following are the most prevailing scripts that not provided on devnet by default

- [ ] Anyone can pay: https://github.com/nervosnetwork/rfcs/blob/master/rfcs/0026-anyone-can-pay/0026-anyone-can-pay.md
- [ ] Simple UDT: https://github.com/nervosnetwork/rfcs/blob/master/rfcs/0025-simple-udt/0025-simple-udt.md
- [ ] CoTA: unrevealed
- [ ] Omnilock: https://github.com/nervosnetwork/rfcs/blob/master/rfcs/0042-omnilock/0042-omnilock.md

But they are widely used in dapps, and are usually inevitable in dapp development. It would be quite helpful if they are out-of-the-box in kuai

Daryl-L:
What I designed is as follow.

First, I will provider a method named `depolyInfraScripts` in `CkbDockerNetwork` for cli to call, cli will call this method immediately while the node starting. Then I will create a folder which stores the scripts should be deployed.

In file https://github.com/ckb-js/kuai/blob/11d213db47e9ba025e209ea26dbe612ea5967604/packages/core/src/builtin-tasks/node.ts#L27

```typescript
const ckbDockerNetwork = new CkbDockerNetwork()
await ckbDockerNetwork.start({
  port: port.toString(),
  detached,
  genesisAccountArgs: genesisArgs,
})
await ckbDockerNetwork.depolyInfraScripts()
```

Create a folder to store scripts in package `docker-node`.
```
- docker-node
- - scripts
- - - sudt
- - - omnilock
- - - acp
```

In file https://github.com/ckb-js/kuai/blob/11d213db47e9ba025e209ea26dbe612ea5967604/packages/docker-node/src/ckb-docker-network.ts#L9
```typescript
deploy() {
  // load scripts bin from scripts file
  // deploy them by prepared account
  // output the cell info of deployed scripts to command line and file
```


What I inspires me is Force Bridge. https://github.com/Magickbase/force-bridge/blob/feature-nervos-to-ethereum/offchain-modules/packages/x/src/ckb/tx-helper/deploy.ts
Please correct me if I was wrong.

Keith-CY:
> What I designed is as follow.
> 
> First, I will provider a method named `depolyInfraScripts` in `CkbDockerNetwork` for cli to call, cli will call this method immediately while the node starting. Then I will create a folder which stores the scripts should be deployed.
[...]

PR #257, opened 2023-05-05
feat(dev-scripts): add deployment for dev scripts

Daryl-L:
1. add some scripts which are helpful in development
2. use the ckb bin to initiate nodes instead of using docker

#235


Keith-CY:
How is it going

packages/core/src/builtin-tasks/node.ts
```diff
@@ -4,20 +4,28 @@ import { CkbDockerNetwork } from '@ckb-js/kuai-docker-node'
...
+  filePath: string
 }
 
 const TERMINATION_SIGNALS = ['SIGINT', 'SIGTERM', 'SIGQUIT']
+const BUILTIN_SCRIPTS = ['anyone_can_pay', 'omni_lock', 'simple_udt']
 
 subtask('node:start', 'start a ckb node')
   .addParam('port', 'The port of the node', 8114, paramTypes.number)
   .addParam('detached', 'Run the node in detached mode', false, paramTypes.boolean)
+  .addParam('filePath', 'The script config file path', undefined, paramTypes.string, true, true)
```
Keith-CY, line 26: The `param` name could be more intuitive to indicate users it's about `script`, or it will be treated as `ckb node configuration`, e.g., `ckb.toml`

packages/core/src/builtin-tasks/node.ts
```diff
@@ -31,9 +39,31 @@ subtask('node:start', 'start a ckb node')
...
-    env.config.network = {
-      url: 'http://localhost:' + port,
+    env.config.network = { url: ckbDockerNetwork.url }
+
+    const builtInDirPath = cachePath('built-in')
+    for (const script of BUILTIN_SCRIPTS) {
+      const filePath = path.join(builtInDirPath, script)
+      if (!fs.existsSync(filePath)) {
+        download(
+          `https://github.com/Daryl-L/ckb-production-scripts/releases/download/scripts/${script}`,
```
Keith-CY, line 49: It's magic, better to be a constant in configuration

packages/core/src/builtin-tasks/node.ts
```diff
@@ -31,9 +39,31 @@ subtask('node:start', 'start a ckb node')
...
+          builtInDirPath,
+          { filename: script },
+        )
+      }
     }
+    await new Promise((resolve) => setTimeout(resolve, 20000))
```
Keith-CY, line 55: `timersPromises.scheduler.wait(delay[, options])` or `timersPromises.scheduler.yield()` can be used as `sleep`

Ref: https://nodejs.org/api/timers.html#timerspromisesschedulerwaitdelay-options

packages/core/src/builtin-tasks/node.ts
```diff
@@ -31,9 +39,31 @@ subtask('node:start', 'start a ckb node')
...
+    await ckbDockerNetwork.generateLumosConfig()
+    config.initializeConfig(ckbDockerNetwork.lumosConfig)
+
+    await ckbDockerNetwork.deployScripts({
+      builtInScriptName: BUILTIN_SCRIPTS,
+      configFilePath: filePath ?? path.resolve(configPath(), 'scripts.json'),
+      builtInDirPath,
+      indexer: new Indexer(`${ckbDockerNetwork.url}/indexer`),
+      rpc: new RPC(`${ckbDockerNetwork.url}/rpc`),
+      privateKey: '0xfd686a48908e8caf97723578bf85f746e1e1d8956cb132f6a2e92e7234a2a245',
```
Keith-CY, line 65: How is the private key from

packages/core/src/builtin-tasks/node.ts
```diff
@@ -31,9 +39,31 @@ subtask('node:start', 'start a ckb node')
...
 
-    env.config.network = {
-      url: 'http://localhost:' + port,
+    env.config.network = { url: ckbDockerNetwork.url }
+
+    const builtInDirPath = cachePath('built-in')
+    for (const script of BUILTIN_SCRIPTS) {
+      const filePath = path.join(builtInDirPath, script)
+      if (!fs.existsSync(filePath)) {
+        download(
```
Keith-CY, line 48: How about using the official network client [undici](https://github.com/nodejs/undici) which will be a built-in module

packages/core/src/helper.ts
```diff
@@ -16,3 +19,19 @@ export async function initialKuai(args: KuaiArguments = {}): Promise<KuaiContext
...
+  return cachePath
+}
+
+export const configPath = (...paths: string[]) => {
+  const configPath = path.resolve(PATH.config, ...paths)
+  if (!fs.existsSync(configPath)) {
+    fs.mkdirSync(configPath, { recursive: true })
+  }
+  return configPath
+}
```
Keith-CY, line 37: The logic of two methods are same so that they can be curried

packages/docker-node/src/ckb-docker-network.ts
```diff
@@ -1,13 +1,18 @@
...
+import { Address, Indexer, RPC, commons, helpers, hd, Transaction, config } from '@ckb-lumos/lumos'
 
 const CKB_NODE_IMAGE = 'kuai/ckb-dev'
 const DOCKER_SOCKET_PATH = process.env.DOCKER_SOCKET || '/var/run/docker.sock'
 
 export class CkbDockerNetwork {
   private _docker: Docker
+  private _lumosConfig: config.Config = config.getConfig()
+  private _port = '8114'
+  private _host = 'localhost'
```
Keith-CY, line 15: They can be defined as `private field`

Ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields

packages/docker-node/src/ckb-docker-network.ts
```diff
@@ -63,6 +69,110 @@ lock.hash_type = "type"\n`
...
+    privateKey: string
+  }): Promise<void> {
+    const from = helpers.encodeToConfigAddress(hd.key.privateKeyToBlake160(privateKey), 'SECP256K1_BLAKE160')
+    await indexer.waitForSync()
+    const config = {
+      builtIn: await this.deployBuiltInScripts(builtInScriptName, builtInDirPath, indexer, rpc, from, privateKey),
+      custom: await this.deployCustomScripts(configFilePath, indexer, rpc, from, privateKey),
+    }
+
+    console.log(configFilePath)
```
Keith-CY, line 172: `log` could be `info` or `debug`

packages/docker-node/src/ckb-docker-network.ts
```diff
@@ -63,6 +69,110 @@ lock.hash_type = "type"\n`
...
+      const configs = JSON.parse(fs.readFileSync(filePath).toString())
+      if (configs.custom && Array.isArray(configs.custom)) {
+        for (const script of configs.custom) {
+          scripts.push(await this.doDeploy(rpc, indexer, from, privateKey, script, script.path))
+        }
+      }
+    }
+
+    return scripts
+  }
```
Keith-CY, line 148: How about deploying built-in scripts via the workflow for custom scripts so there won't be two methods to deploy a script.
Specifically, why are `scriptName: string[]` required in `deployBuiltInScripts`

packages/core/src/builtin-tasks/node.ts
```diff
@@ -31,9 +39,31 @@ subtask('node:start', 'start a ckb node')
       genesisAccountArgs: genesisArgs,
     })
 
-    env.config.network = {
-      url: 'http://localhost:' + port,
+    env.config.network = { url: ckbDockerNetwork.url }
```
PainterPuppets, line 42: Rebase the latest `develope`, there seems to be a conflict here

packages/core/src/builtin-tasks/node.ts
PainterPuppets, line 65: Here can add to the default config

packages/core/src/builtin-tasks/node.ts
```diff
@@ -31,9 +39,31 @@ subtask('node:start', 'start a ckb node')
...
     }
+    await new Promise((resolve) => setTimeout(resolve, 20000))
+    await ckbDockerNetwork.generateLumosConfig()
+    config.initializeConfig(ckbDockerNetwork.lumosConfig)
+
+    await ckbDockerNetwork.deployScripts({
+      builtInScriptName: BUILTIN_SCRIPTS,
+      configFilePath: filePath ?? path.resolve(configPath(), 'scripts.json'),
+      builtInDirPath,
+      indexer: new Indexer(`${ckbDockerNetwork.url}/indexer`),
```
PainterPuppets, line 63: Now that the rpc url integrates the indexer, you can create an indexer directly with the rpc url

packages/core/src/builtin-tasks/node.ts
```diff
@@ -31,9 +39,31 @@ subtask('node:start', 'start a ckb node')
...
+        )
+      }
     }
+    await new Promise((resolve) => setTimeout(resolve, 20000))
+    await ckbDockerNetwork.generateLumosConfig()
+    config.initializeConfig(ckbDockerNetwork.lumosConfig)
+
+    await ckbDockerNetwork.deployScripts({
+      builtInScriptName: BUILTIN_SCRIPTS,
+      configFilePath: filePath ?? path.resolve(configPath(), 'scripts.json'),
```
PainterPuppets, line 61: The configuration should be taken from `env.config`, in which case it can be configured in the user's kuai.config.ts

packages/core/src/builtin-tasks/node.ts
Daryl-L: > How is the private key from

It is from the file `dev-template.toml`

packages/docker-node/src/ckb-docker-network.ts
Daryl-L: > They can be defined as `private field`
> 
> Ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields

I just want to keep them the same as what other private property defined before, like `_docker`. Or I will change the `_docker` to `#docker` in this PR?

packages/core/src/builtin-tasks/node.ts
Keith-CY: It would be good to have a reference to the origin so this info won't be lost

packages/docker-node/src/ckb-docker-network.ts
Keith-CY: `#docker` would be better because it's native JavaScript while `private` is `TypeScript` and not forcely restrcited.

yanguoyu (approved):

packages/docker-node/src/types/index.ts
```diff
@@ -1,5 +1,13 @@
...
 export interface DockerNodeStartOptions {
   port: string
   detached?: boolean
   genesisAccountArgs?: string[]
 }
+
+export interface InfraScript {
+  name: string
+  path: string
+  cellDep: { cellDeps?: CellDep[] } & CellDep
```
Keith-CY, line 12: The type declaration is hard to understand, it will be 

```typescript
{
  cellDeps: CellDep[],
  ...CellDep,
}
```

Keith-CY: And only `cellDep: CellDep` is used at https://github.com/ckb-js/kuai/pull/257/files#diff-82a2e05889db2d2dbb18f524baf5085533ddf8c66a15de72075b77e2d2fc00e6R104

What's the use cases of `CellDeps` and `...CellDep`

packages/docker-node/src/ckb-docker-network.ts
Daryl-L: > How about deploying built-in scripts via the workflow for custom scripts so there won't be two methods to deploy a script. Specifically, why are `scriptName: string[]` required in `deployBuiltInScripts`

I have extract the logic to the method `doDeploy`, and I think it is conveniently for other modification by separating them to two methods.

packages/docker-node/src/types/index.ts
Daryl-L: > And only `cellDep: CellDep` is used at https://github.com/ckb-js/kuai/pull/257/files#diff-82a2e05889db2d2dbb18f524baf5085533ddf8c66a15de72075b77e2d2fc00e6R104
> 
> What's the use cases of `CellDeps` and `...CellDep`

`...CellDep` is used to recording the out point information of the script which is deployed, and the `cellDeps` is used to providing the dependencies of the script which will be deployed.

packages/docker-node/src/ckb-docker-network.ts
Keith-CY: The point I want to deliver is that `custom` and `built-in` scripts can be treated in the same manner. They're all binaries to be loaded.

Here we have two methods to load binaries, one from **built-in file path**, one from **config.custom**

What if we preset `built-in scripts` in the generated `config.builtIn`.

In fact, `config.custom` can be generalized as `config.scripts` with `built-in scripts` as its default value. Custom scripts can be appended to `config.scripts`.

But we can leave it now, because it's just about preset **explicit** or **implicit**

packages/docker-node/src/types/index.ts
Keith-CY: Seems that `CellDep` can be treated as `script's outPoint` or something. Putting `CellDep` and `CellDeps` together is hard to understand their relationship.

Keith-CY:
How is it going

packages/docker-node/src/types/index.ts
Keith-CY: Will we update this type declaration?

Daryl-L: > Will we update this type declaration?

Yes, I will push the commit later after test

packages/core/src/builtin-tasks/node.ts
```diff
@@ -30,6 +41,38 @@ subtask('node:start', 'start a ckb node')
...
+    config.initializeConfig(ckbDockerNetwork.lumosConfig)
+
+    await ckbDockerNetwork.deployScripts({
+      builtInScriptName: BUILTIN_SCRIPTS,
+      configFilePath: env.config.devNode?.builtInContractConfigPath ?? path.resolve(configPath(), 'scripts.json'),
+      builtInDirPath,
+      indexer: new Indexer(ckbDockerNetwork.url),
+      rpc: new RPC(ckbDockerNetwork.url),
+      privateKey: env.config.kuaiArguments.privateKey ?? DEFAULT_KUAI_PRIVATE_KEY,
+    })
```
PainterPuppets, line 75: I was thinking that the deployment of these contracts should be a function maintained by the `ckb node` itself, and that different `ckb nodes` might have different ways of deploying built-in contracts, similar to the way `genesisAccountArgs` is passed above

BTW, maybe we should abstract `ckb node` into an interface, so that if we want to start `ckb node` in other ways, we can just implement the corresponding interface

Keith-CY: >  that different `ckb nodes` might have different ways of deploying built-in contracts

Could you elaborate on the reason that different node types will have different progress to deploy built-in contracts?

In my mind, the progress can be abstracted into the following steps:

1. build contract source code into binaries
2. send the binaries one by one to the target node

There's no keyword of node type bifurcates the workflow.

> BTW, maybe we should abstract ckb node into an interface, so that if we want to start ckb node in other ways, we can just implement the corresponding interface

+1

Daryl-L: > I was thinking that the deployment of these contracts should be a function maintained by the `ckb node` itself, and that different `ckb nodes` might have different ways of deploying built-in contracts, similar to the way `genesisAccountArgs` is passed above
> 
> BTW, maybe we should abstract `ckb node` into an interface, so that if we want to start `ckb node` in other ways, we can just implement the corresponding interface

Yes, I think all these could be done by abstracting an interface. The `ckb docker node` is an implementation of the interface.

Daryl-L: I think we can do this in another branch, which is specifically for this.

Keith-CY (approved):

packages/core/src/builtin-tasks/node.ts
Keith-CY: Please open a new issue for track

packages/docker-node/src/ckb-docker-network.ts
Keith-CY: Tracked by https://github.com/ckb-js/kuai/issues/304

PR #267, opened 2023-05-16
fix: promise order bug

PainterPuppets:
Supplementary 0 2 multisig example 
image

### DIFF
--- a/packages/core/src/builtin-tasks/contract.ts
+++ b/packages/core/src/builtin-tasks/contract.ts
@@ -154,13 +154,13 @@ subtask('contract:sign-message')
         throw new KuaiError(ERRORS.BUILTIN_TASKS.NOT_SPECIFY_SIGNING_ADDRESS)
       }
 
-      const multisigs = await Promise.all(
-        preSigningAddresses.map(async (addr) => {
-          const password = await read({ prompt: `Input ${addr}'s password for sign messge by ckb-cli:`, silent: true })
-          console.info('')
-          return signMessageByCkbCli(message, addr, password).slice(2)
-        }),
-      )
+      let multisigs: string[] = []
+      for (const addr of preSigningAddresses) {
+        const password = await read({ prompt: `Input ${addr}'s password for sign messge by ckb-cli:`, silent: true })
+        console.info('')
+        const sig = signMessageByCkbCli(message, addr, password).slice(2)
+        multisigs = [...multisigs, sig]
+      }
 
       return `${prefix}${multisigs.join('')}`
     }


Daryl-L (approved):

Keith-CY:
The example shows how to deploy contracts by multisig signature on the same machine. The fact is that deployers would be on different platforms so the partially signed transaction should be shared from one to another. Can `kuai` do this now?

PainterPuppets:
> The example shows how to deploy contracts by multisig signature on the same machine. The fact is that deployers would be on different platforms so the partially signed transaction should be shared from one to another. Can `kuai` do this now?
It may depend on the user's signature method

I don't have contact with multi-signer users, and I imagine that they probably sign by both automatic and manual signature collection

Automatically, it could be some kind of rustful api that passes a message and returns a signature

This approach is not a universal solution, each user may have their own signature approach, I think we should use a custom signer to handle it
```ts
// kusi.config.js

const { subtask } = require('@ckb-js/kuai-core')

subtask('contract:sign-message')
  .addParam('message', 'message to be signed', '', paramTypes.string, false)
  .addParam('address', 'the address of message signer', '', paramTypes.string, false)
  .addParam('prefix', 'the prefix of signature', '', paramTypes.string, true)
  .addParam(
    'signer',
    'signer provider [default: ckb-cli] [possible values: ckb-cli, ckb-cli-multisig]',
    'ckb-cli',
    paramTypes.string,
    true,
  )
  .setAction(async ({ message, address, prefix = '', signer }, _, runSuper) => {
    if (signer === 'custom-signer') {
      const signatures = await doSomething(message);
      return `${prefix}${signatures.join('')}`
    }

    if (runSuper.isDefined) {
      await runSuper()
    }
  })


```


If the user's signature is manually spliced, e.g. sending the message to different people via email? And then splice the message by email or something

If so, we can add a multisig-directly signer process is probably, when signing print out the messge to be signed, then the user use this messge to send an email to other people or other methods, collect all the signatures and then fill to kuai?

Do you think these two solutions work?

Keith-CY:
> It may depend on the user's signature method
> 
> I don't have contact with multi-signer users, and I imagine that they probably sign by both automatic and manual signature collection
> 
> Automatically case, it could be some kind of rustful api that passes a message and returns a signature
> 
> This is not a universal case, each user may have their own signature method, I think should use a custom signer to handle it
> 
> ```ts
> // kusi.config.js
> 
> const { subtask } = require('@ckb-js/kuai-core')
> 
> subtask('contract:sign-message')
>   .addParam('message', 'message to be signed', '', paramTypes.string, false)
>   .addParam('address', 'the address of message signer', '', paramTypes.string, false)
>   .addParam('prefix', 'the prefix of signature', '', paramTypes.string, true)
>   .addParam(
>     'signer',
>     'signer provider [default: ckb-cli] [possible values: ckb-cli, ckb-cli-multisig]',
>     'ckb-cli',
>     paramTypes.string,
>     true,
>   )
>   .setAction(async ({ message, address, prefix = '', signer }, _, runSuper) => {
>     if (signer === 'custom-signer') {
>       const signatures = await doSomething(message);
>       return `${prefix}${signatures.join('')}`
>     }
> 
>     if (runSuper.isDefined) {
>       await runSuper()
>     }
>   })
> ```
> 
> If the user's signature from manually, e.g. sending the message to different people via email? And then splice the message by email or something
> 
> If so, we can add a `multisig-directly` signer process is probably, when signing print out the messge to be signed, then the user use this messge to send an email to other people or other methods, collect all the signatures and then fill to kuai?
> 
> Do you think these two solutions work?

I would suggest reusing the workflow adopted in ckb-cli and neuron, that is

1. sign the transaction with 1 signer
2. export the partially signed transaction
3. send the partially signed transaction to another signer
4. sign the partially signed transaction
5. repeat until the transaction is fully signed
6. submi the transaction to deploy the contract

The main steps are
1. export `partially signed transaction` to share
2. import `partially signed transaction` to sign

yanguoyu (approved):

PR #282, opened 2023-05-25
feat: Add type doc to build doc for project.

yanguoyu:
1. Use decorator mode in sample-projects
2. Add build doc after creating a project
3. Add `quick start` in `README`
4. Add `Development.md` for sample-projects

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,3 +31,66 @@ Outputs will be built in the `lib` directory in their own workspaces
 ## Demos
 
 [MVP DApp](./packages/samples/mvp-dapp)
+
+## Quick start
+
+#### Environment
+
+    node >= 18
+
+#### Kuai Installation
+
+First, install `kuai` globally by `npm link`:
+
+```bash
+$ git clone git@github.com:ckb-js/kuai.git
+
+$ cd kuai
+
+$ yarn bootstrap
+
+$ yarn build
+
+$ cd packages/cli
+
+$ npm link
+```
+
+#### Create a template project using the command.
+
+```bash
+kuai init
+```
+
+#### Running
+
+`kuai` has `dev` and `prod` modes
+
+dev
+
+```bash
+npm run dev
+```
+
+prod
+
+```bash
+npm run build
+npm run start:prod
+```
+
+#### Others commands
+
+run test
+
+```bash
+npm run test
+```
+
+build docs
+
+The docs is build by [typedoc](https://typedoc.org/), `typedoc.json` is the build configuration.
+
+```bash
+npm run doc
+```
--- a/packages/core/sample-projects/Development.md
+++ b/packages/core/sample-projects/Development.md
@@ -1,0 +1,26 @@
+# Development
+
+## Environment
+
+You should start a redis service. And the redis's default port is 6379. You can edit it at `kuai.config.ts`
+
+```typescript
+import { KuaiConfig } from '@ckb-js/kuai-core';
+
+const config: KuaiConfig = {
+  REDIS_PORT: 6379,
+};
+
+export default config;
+```
+
+## Quick Start
+
+```sh
+git clone <remote_repository_location>
+cd <project name>
+npm install
+npm run dev
+```
+
+It will start a service at `http://127.0.0.1:3000`, try to open [hello](http://127.0.0.1:3000)
--- a/packages/core/sample-projects/kuai.config.ts
+++ b/packages/core/sample-projects/kuai.config.ts
@@ -2,6 +2,10 @@ import { KuaiConfig } from '@ckb-js/kuai-core';
 
 const config: KuaiConfig = {
   port: 3000,
+  ckbChain: {
+    rpcUrl: 'http://127.0.0.1:8114',
+    prefix: 'ckt',
+  },
 };
 
 export default config;
--- a/packages/core/recommended-package.json
+++ b/packages/core/recommended-package.json
@@ -5,6 +5,7 @@
     "build": "tsc",
     "build:watch": "tsc -w",
     "start:prod": "node ./dist/main.js",
-    "test": "jest"
+    "test": "jest",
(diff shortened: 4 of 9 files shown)


Daryl-L (approved):

README.md
```diff
@@ -31,3 +31,64 @@ Outputs will be built in the `lib` directory in their own workspaces
 ## Demos
 
 [MVP DApp](./packages/samples/mvp-dapp)
+
+## Quick start
+
+#### Enviroment
```
Keith-CY, line 37:
```suggestion
#### Environment
```

README.md
```diff
@@ -31,3 +31,64 @@ Outputs will be built in the `lib` directory in their own workspaces
...
+
+    node > 18
```
Keith-CY, line 39:
```suggestion
    node >= 18
```

README.md
```diff
@@ -31,3 +31,64 @@ Outputs will be built in the `lib` directory in their own workspaces
...
+
+First, you need to install `kuai` globally using `npm` or `yarn`, or just use `npx`:
+
+```bash
+# Use npm to install CLI
+$ npm install -g kuai
+
+# OR Use yarn to install CLI
+$ yarn global add kuai
+```
```
Keith-CY, line 51: `npx` is prior to global installation because it won't affect global environment, and always update-to-date.

packages/core/sample-projects/Development.md
```diff
@@ -0,0 +1,25 @@
+# Development
+
+## Environment
+
+You need to start a redis service. And the redis's default port is 6379. You can edit it at `kuai.config.ts`
```
Keith-CY, line 5: `Subject` in sentences should be the service/program instead of `human`.
`You need`, `You should` sound like commands, which is not polite.

In some cases, `I think`, `I figure` sound not objective.

packages/core/sample-projects/Development.md
```diff
@@ -0,0 +1,25 @@
...
+  REDIS_PORT: 6379,
+};
+
+export default config;
+```
+
+## Quick Start
+
+```sh
+git clone
```
Keith-CY, line 20: Is `git clone` a complete command here? And is there a `cd` command after `git clone`?

README.md
yanguoyu: If the developer uses `npx`, it is unnecessary to install globally. Then they can skip this step.

packages/core/sample-projects/Development.md
yanguoyu: Because it depends on the new project name, it will fill with the project name that creates by `kuai init`. 
It may need to be improved later by a template.

Keith-CY: Placeholder as `<project name>` could be used to imply the parameter, similar to 
```
# To download a file:
curl <url>

# To download and rename a file:
curl <url> -o <outfile>

# To download multiple files:
curl -O <url> -O <url>

# To download all sequentially numbered files (1-24):
curl http://example.com/pic[1-24].jpg

# To download a file and pass HTTP authentication:
curl -u <username>:<password> <url>

# To download a file with a proxy:
curl -x <proxy-host>:<port> <url>

# To download a file over FTP:
curl -u <username>:<password> -O ftp://example.com/pub/file.zip

# To get an FTP directory listing:
curl ftp://username:password@example.com

# To resume a previously failed download:
curl -C - -o <partial-file> <url>

# To fetch only the HTTP headers from a response:
curl -I <URL>
```

packages/core/sample-projects/kuai.config.ts
```diff
@@ -2,6 +2,10 @@ import { KuaiConfig } from '@ckb-js/kuai-core';
 
 const config: KuaiConfig = {
   port: 3000,
+  ckbChain: {
```
PainterPuppets, line 5: `ckbChain` will be obtained through the `network` and `networks` values, so it is more expected to set these two instead of setting `ckbChain` directly, in addition there are some built-in network config, here can set it to `{ network: 'devnet' }` to achieve the same goal

yanguoyu: From https://github.com/ckb-js/kuai/blob/develop/packages/core/src/type/runtime.ts#L15, the `ckbChain` is necessary, maybe we can set it as optional?

If I don't define `ckbChain`, it will throw an exception when running `yarn dev` or `yarn build`.

README.md
```diff
@@ -31,3 +31,64 @@ Outputs will be built in the `lib` directory in their own workspaces
...
+
+First, install `kuai` globally using `npm` or `yarn`, or just use `npx`:
+
+```bash
+# Use npm to install CLI
+$ npm install -g kuai
+
+# OR Use yarn to install CLI
+$ yarn global add kuai
+```
```
Keith-CY, line 51: After a second thought, I found this section is not available now, because 
1. `kuai-cli` is not published yet
2. `kuai-cli` won't be named `kuai` directly, it should have a npm scope, as coded at https://github.com/ckb-js/kuai/blob/develop/packages/cli/package.json#L2

I would suggest adding `npm link` here

Ref: https://docs.npmjs.com/cli/v8/commands/npm-link

packages/core/sample-projects/Development.md
```diff
@@ -0,0 +1,26 @@
...
+  REDIS_PORT: 6379,
+};
+
+export default config;
+```
+
+## Quick Start
+
+```sh
+git clone remote_repository_location
```
Keith-CY, line 20:
```suggestion
git clone <remote repository location>
```

Keith-CY (approved):

packages/core/sample-projects/kuai.config.ts
Keith-CY: Any update? @PainterPuppets

PainterPuppets (approved):

PR #288, opened 2023-05-30
fix(config): remove default network config set in argument

Daryl-L:
I think, the default network config should not be set into argument. The default network config should only be set when the network config is being used


yanguoyu:
Test failed.

PainterPuppets (approved):

yanguoyu (approved):

PR #305, opened 2023-06-08
feat: use contract with kuai

yanguoyu:


packages/samples/mvp-dapp/src/actors/record.model.ts
```diff
@@ -42,8 +50,10 @@ export type StoreType = {
  */
 @ActorProvider({ ref: { name: 'record', path: '/:codeHash/:hashType/:args/' } })
 @LockPattern()
+@TypePattern()
 @DataPrefixPattern(DAPP_DATA_PREFIX)
```
Keith-CY, line 54: Is the `DataPrefixPattern` redundant now?

yanguoyu: It uses to filter cell's data that start with `DAPP_DATA_PREFIX`

Keith-CY: But `TypePattern` is added above, isn't it enough to filter cells of the `mvp contract`?

yanguoyu: You are right, it's enough.

Issue #312, opened 2023-06-13
It might not be able to use kuai to start a node on Apple M chip platform


Daryl-L:
I have learnt from Acceptance that the ckb could not run in docker with Apple M1, as the kuai is using docker to start a node, so it is not supported.

Keith-CY:
> I have learnt from Acceptance that the ckb could not run in docker with Apple M1, as the kuai is using docker to start a node, so it is not supported.

I'll check it with the core team for a solution

Keith-CY:
Error Message

```bash
<jemalloc>: MADV_DONTNEED does not work (memset will be used instead)
<jemalloc>: (This is the expected behaviour if you are running under QEMU)

Error: read ETIMEDOUT
    at TLSWrap.onStreamRead (node:internal/stream_base_commons:217:20) {
  errno: -60,
  code: 'ETIMEDOUT',
  syscall: 'read'
}
```

Keith-CY:
Check with the core team how long it will take

Daryl-L:
#319

PR #319, opened 2023-06-16
feat(node): abstract the ckb node interface

Daryl-L:
1. abstract the ckb node interface
2. add an implementation for starting ckb node with binary
3. add an implementation for existence ckb node

#312


packages/docker-node/src/ckb-bin-network.ts
```diff
@@ -0,0 +1,259 @@
...
+
+export class CKBBinNetwork implements CKBNode {
+  #port = '8114'
+  #host = '127.0.0.1'
+  #lumosConfig: config.Config = config.getConfig()
+
+  stop({ ckbPath, clear }: BinNodeStopOptions): void {
+    if (fs.existsSync(path.resolve(ckbPath, 'pid'))) {
+      const indexer = fs.readFileSync(path.resolve(ckbPath, 'pid', 'indexer'), 'utf-8')
+      spawn('kill', ['-9', indexer])
```
yanguoyu, line 18: We'd better not stop ckb by `kill -9`, it may break the ckb data file, then it cannot start next time.
It's better to `kill 15` means `kill with SIGTERM`. https://zh.wikipedia.org/wiki/Unix%E4%BF%A1%E5%8F%B7#SIGTERM

packages/docker-node/src/ckb-bin-network.ts
```diff
@@ -0,0 +1,259 @@
...
+
+      fs.copyFileSync(path.resolve(__dirname, '../ckb/ckb-miner.toml'), path.join(ckbPath, 'ckb-miner.toml'))
+    }
+
+    if (!fs.existsSync(path.join(ckbPath, 'pid'))) {
+      fs.mkdirSync(path.join(ckbPath, 'pid'))
+    }
+  }
+
+  async #startIndexer(ckbPath: string, detached: boolean) {
```
yanguoyu, line 56: IMO, we can add an indexer parameter to start the node with indexer or without indexer
```suggestion
  async #startNode(ckbPath: string, detached: boolean, indexer: boolean = true) {
```

packages/docker-node/src/interface.ts
```diff
@@ -0,0 +1,15 @@
...
+  url: string
+  port: string
+  host: string
+  lumosConfig: config.Config
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  start(params: any): void
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  stop(params: any): void
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  deployScripts(params: any): void
```
yanguoyu, line 13: Could we provide some necessary parameters for these functions?

packages/docker-node/src/ckb-bin-network.ts
```diff
@@ -0,0 +1,259 @@
...
+    try {
+      await waitUntilCommitted(rpc, txHash)
+    } catch (e) {
+      console.error(e)
+    }
+
+    return {
+      name: script,
+      path: filePath,
+      depType: 'code',
```
yanguoyu, line 133: Is the `depType` necessarily `code`?

packages/core/src/builtin-tasks/node.ts
```diff
@@ -23,64 +25,122 @@ const TERMINATION_SIGNALS = ['SIGINT', 'SIGTERM', 'SIGQUIT']
 const BUILTIN_SCRIPTS = ['anyone_can_pay', 'omni_lock', 'simple_udt']
 const DEFAULT_BUILTIN_CONTRACT_DOWNLOAD_BASE_URL =
   'https://github.com/ckb-js/ckb-production-scripts/releases/download/scripts'
+const DEFAULT_CKB_BIN_DOWNLOAD_BASE_URL = 'https://github.com/nervosnetwork/ckb/releases/download'
+const DEFAULT_CKB_BIN_VERSION = 'v0.109.0'
```
Keith-CY, line 29: The latest version can be fetched by

```bash
curl -L \
  -H "Accept: application/vnd.github+json" \
  https://api.github.com/repos/nervosnetwork/ckb/releases/latest
```

packages/core/src/builtin-tasks/node.ts
```diff
@@ -23,64 +25,122 @@ const TERMINATION_SIGNALS = ['SIGINT', 'SIGTERM', 'SIGQUIT']
...
+
+const startBinNode = async (version: string, port: number, genesisArgs: string[]): Promise<CKBNode> => {
+  const packageName = `ckb_${version}_${os.machine()}-${osPlatform()}`
+  const packageFileName = `${packageName}.${packageType()}`
+
+  if (!fs.existsSync(path.resolve(cachePath('ckb', 'bin'), packageName))) {
```
Keith-CY, line 35: The cached binary path is referenced times and can be assigned to a variable

packages/core/src/builtin-tasks/node.ts
```diff
@@ -23,64 +25,122 @@ const TERMINATION_SIGNALS = ['SIGINT', 'SIGTERM', 'SIGQUIT']
...
+    await download(
+      `${DEFAULT_CKB_BIN_DOWNLOAD_BASE_URL}/${version}/${packageFileName}`,
+      path.resolve(cachePath('ckb', 'zip'), `${packageFileName}`),
+    )
+
+    if (osPlatform() === 'unknown-linux-gnu') {
```
Keith-CY, line 41: I'm not sure if the condition is proper. I would use a condition of if `tar` or `unzip` exists.

packages/core/src/builtin-tasks/node.ts
```diff
@@ -23,64 +25,122 @@ const TERMINATION_SIGNALS = ['SIGINT', 'SIGTERM', 'SIGQUIT']
...
+      execSync.execSync(
```
Keith-CY, line 42: What if importing `execSync` directly

packages/core/src/builtin-tasks/node.ts
Daryl-L: > The latest version can be fetched by
> 
> ```shell
> curl -L \
>   -H "Accept: application/vnd.github+json" \
>   https://api.github.com/repos/nervosnetwork/ckb/releases/latest
> ```

I see, I will change the way for getting the package information by using this api.

packages/core/src/builtin-tasks/node.ts
Daryl-L: > I'm not sure if the condition is proper. I would use a condition of if `tar` or `unzip` exists.

I think I would better to checking whether the extract tool exists or not first. If not, I will print an error message.

Keith-CY:
Is this PR ready for review?

Daryl-L:
> Is this PR ready for review?

Yes, and 2 comments are still being resolved.

packages/docker-node/src/types/index.ts
```diff
@@ -1,9 +1,30 @@
...
   detached?: boolean
-  genesisAccountArgs?: string[]
+}
+
+export interface DeployOptions {
+  builtInScriptName: string[]
+  configFilePath: string
+  builtInDirPath: string
+  indexer: Indexer
+  rpc: RPC
```
PainterPuppets, line 14: The ckb node should know where the rpc endpoint is when deploying the contract, and adding the `lumos` dependency at the interface level feels a bit strange, and how the rpc is called should be put into the concrete implementation. 
Assuming we have a virtual node like evm, we may not use such an rpc to interact

packages/docker-node/src/interface.ts
```diff
@@ -0,0 +1,15 @@
+import { config } from '@ckb-lumos/lumos'
+
+export interface CKBNode {
+  url: string
+  port: string
+  host: string
+  lumosConfig: config.Config
```
PainterPuppets, line 7: Why does ckb node's interface depend on `lumos`

yanguoyu (approved):

packages/core/src/builtin-tasks/node.ts
```diff
@@ -23,64 +24,124 @@ const TERMINATION_SIGNALS = ['SIGINT', 'SIGTERM', 'SIGQUIT']
 const BUILTIN_SCRIPTS = ['anyone_can_pay', 'omni_lock', 'simple_udt']
 const DEFAULT_BUILTIN_CONTRACT_DOWNLOAD_BASE_URL =
   'https://github.com/ckb-js/ckb-production-scripts/releases/download/scripts'
+const DEFAULT_CKB_BIN_DOWNLOAD_BASE_URL = 'https://github.com/nervosnetwork/ckb/releases/download'
+const DEFAULT_CKB_BIN_VERSION = 'v0.110.0'
```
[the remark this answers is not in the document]
Daryl-L: I have changed the way to fetch the latest version by using API. And keep this hard code is for the condition if the version could not be fetched such as the API is not usable by fault.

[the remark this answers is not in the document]
Keith-CY: It brings a mental burden to update it manually and timely. What if loading the default version from a file that is expected to be updated on publish, and throws when fails to be set?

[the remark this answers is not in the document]
Daryl-L: > It brings a mental burden to update it manually and timely. What if loading the default version from a file that is expected to be updated on publish, and throws when fails to be set?

OK, I will have a try.

packages/docker-node/src/interface.ts
Daryl-L: > Could we provide some necessary parameters for these functions?

Each implementation of the `Node` may has different parameters to start or stop, so I use the `any` as the parameter type here.

packages/docker-node/src/types/index.ts
Daryl-L: These property is not the dependencies for the `Node`, they provider the Lumos config which is from the `Node`.

packages/docker-node/src/interface.ts
Daryl-L: These property is not the dependencies for the Node, they provider the Lumos config which is from the Node.

packages/common/src/util.ts
```diff
@@ -1,6 +1,13 @@
 import type { TransactionWithStatus } from '@ckb-lumos/base'
 import type { RPC } from '@ckb-lumos/lumos'
 import { scheduler } from 'node:timers/promises'
+import path from 'node:path'
+import fs from 'fs'
+import type { URL, UrlObject } from 'url'
```
Keith-CY, line 6:
```suggestion
import type { URL, UrlObject } from 'node:url'
import { scheduler } from 'node:timers/promises'
import path from 'node:path'
import fs from 'node:fs'
```

packages/docker-node/src/helper.ts
```diff
@@ -0,0 +1,26 @@
...
+  const extraConfig = genesisAccountArgs.reduce((a, b) => a + generateGenesisAccount(b), '')
+  const config = fs.readFileSync(join(__dirname, '../ckb', 'dev-template.toml'), 'utf-8')
+
+  return config + extraConfig
+}
+
+export const CKBLatestBinVersion = async (): Promise<string | undefined> => {
+  const { body } = await request('https://api.github.com/repos/nervosnetwork/ckb/releases/latest', {
+    method: 'GET',
+    headers: { Accept: 'application/vnd.github+json', 'User-Agent': 'CKB-Js-Kuai-DApp' },
```
Keith-CY, line 21: What does the custom user agent do in this function

Daryl-L: I found it in the GitHub document, if I didn't add this, 403 error would be responded from GitHub api.

Keith-CY: That weird

```bash
> await fetch('https://api.github.com/repos/nervosnetwork/ckb/releases/latest').then(res => res.json()).then(res => res.tag_name)
# 'v0.110.0'
```

works in Node.js@18.16.0

Keith-CY: And if it's specified in github document, it's better to append the link to it as an annotation

Daryl-L: > That's weird
> 
> 
> 
> ```bash
> 
> > await fetch('https://api.github.com/repos/nervosnetwork/ckb/releases/latest').then(res => res.json()).then(res => res.tag_name)
> 
> # 'v0.110.0'
> 
> ```
> 
> 
> 
> works in Node.js@18.16.0

Too weird. I will have a try again rest of day and if it does not work either, I will add the url of document as the annotion.

packages/docker-node/src/helper.ts
```diff
@@ -0,0 +1,28 @@
...
+  const extraConfig = genesisAccountArgs.reduce((a, b) => a + generateGenesisAccount(b), '')
+  const config = fs.readFileSync(join(__dirname, '../ckb', 'dev-template.toml'), 'utf-8')
+
+  return config + extraConfig
+}
+
+export const CKBLatestBinVersion = async (): Promise<string | undefined> => {
+  const { body } = await request('https://api.github.com/repos/nervosnetwork/ckb/releases/latest', {
+    method: 'GET',
+    // Here is why we need to set User-Agent:
```
Keith-CY, line 21:
```suggestion
    // Here is why we need to set a custom User-Agent:
```

Keith-CY (approved):

packages/common/src/util.ts
```diff
@@ -21,3 +28,36 @@ export const waitUntilCommitted = async (rpc: RPC, txHash: string, timeout = 120
...
+          fileStream.close()
+          res(0)
+        })
+        .on('error', (error) => {
+          rej(error)
+        })
+    })
+  } catch (error) {
+    console.error('Error downloading file:', error)
+  }
```
Keith-CY, line 62: It can be simplified as

```javascript
import { pipeline } from 'node:stream/promises'
import fs from 'node:fs'

pipeline(
  await fetch(URL).then(res => res.body),
  fs.createReadableStream(filePath),
).catch(err => console.error('Error on downloading file: ', error)
```

Ref: 
- Difference between `pipeline` and `pipe`: https://stackoverflow.com/questions/58875655/whats-the-difference-between-pipe-and-pipeline-on-streams

PainterPuppets (approved):

packages/docker-node/src/interface.ts
Keith-CY: Can we resolve this conversation?

Daryl-L: > Can we resolve this conversation?

I have no idea this time, may I can have a consideration and make it possible later in another PR.

PR #324, opened 2023-06-18
feat: update mvp-dapp by the deployed contract

Keith-CY:
Most codes are copied from @yanguoyu's work at https://github.com/ckb-js/kuai/pull/305/commits/2d9150845b3d3371fa141d0d937c791b672fb9a1

A convention is introduced in this PR that deployment info of the contract module should be located at `contract/deployed/`(contract/deployed_demo in this case for demonstration). It's updated manually in this demo, but should be generated automatically in the future.

---

Demo preview
https://www.youtube.com/watch?v=-R3EQxcWJVU

### DIFF
--- a/packages/models/src/utils/decorator/provider.ts
+++ b/packages/models/src/utils/decorator/provider.ts
@@ -87,6 +87,15 @@ export const LockPattern = (): ClassDecorator =>
       value.cell.cellOutput.lock.hashType === obj.lockScript.hashType,
   })
 
+// FIXME: refactor with LockPattern
+export const TypePattern = (): ClassDecorator =>
+  Pattern({
+    cellPattern: (obj: { typeScript: Script }) => (value: UpdateStorageValue) =>
+      value.cell.cellOutput.type?.args === obj.typeScript.args &&
+      value.cell.cellOutput.type?.codeHash === obj.typeScript.codeHash &&
+      value.cell.cellOutput.type?.hashType === obj.typeScript.hashType,
+  })
+
 export const DataPattern = (data: string): ClassDecorator =>
   Pattern({ cellPattern: () => (value: UpdateStorageValue) => value.cell.data == data })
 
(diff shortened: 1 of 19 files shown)


yanguoyu (approved):

packages/models/src/utils/decorator/provider.ts
```diff
@@ -107,6 +116,21 @@ export const Lock =
...
+      ProviderKey.TypePattern,
+      (ref?: ActorRef) => {
+        const codeHash = script?.codeHash ?? ref?.params?.codeHash
+        const hashType = script?.hashType ?? ref?.params?.hashType
+        const args = script?.args ?? ref?.params?.args ?? '0x'
+        return codeHash && hashType ? { codeHash, hashType, args } : undefined
+      },
+      target,
+    )
+
```
Daryl-L, line 133: Isn't it better to extract the code like this?

```typescript
const Script =
  (patternType: symbol, script?: Partial<Script>): ClassDecorator =>
  (target: object) =>
    Reflect.defineMetadata(
      patternType,
      (ref?: ActorRef) => {
        const codeHash = script?.codeHash ?? ref?.params?.codeHash
        const hashType = script?.hashType ?? ref?.params?.hashType
        const args = script?.args ?? ref?.params?.args ?? '0x'
        return codeHash && hashType ? { codeHash, hashType, args } : undefined
      },
      target,
    )

export const Type = (script?: Partial<Script>): ClassDecorator => Script(ProviderKey.TypePattern, script)

export const Lock = (script?: Partial<Script>): ClassDecorator => Script(ProviderKey.LockPattern, script)
```

yanguoyu: That will be better, it needs to refactor to that.

Keith-CY: Sure, it's been marked with
```
// FIXME: refactor with Lock
```
It's not done in this PR because of the time left for demonstration

PR #341, opened 2023-06-28
feat: add contract upgrade cli

PainterPuppets:
```bash
Usage: kuai contract upgrade [options]

Options:
  --name [string]            name of the contract to be deployed (default: "")
  --bin-path [path]          path of contract bin file (default: "")
  --migration-path <string>  path of migration file
  --feepayer <string...>     address or multisig config of the transaction fee payer
                             (default: "")
  --deployer <string...>     address or multisig config of the contract deployer
                             (default: "")
  --signer [string]          signer provider [default: ckb-cli] [possible values:
                             ckb-cli, ckb-cli-multisig] (default: "")
  --fee-rate [number]        per transaction's fee, deployment may involve more than
                             one transaction. default: [1000] shannons/Byte
                             (default: 1000)
  --export [path]            export transaction to file (default: "")
  --send                     send transaction directly (default: false)
```

### DIFF
--- a/packages/core/src/builtin-tasks/contract.ts
+++ b/packages/core/src/builtin-tasks/contract.ts
@@ -31,7 +31,6 @@ interface ContractDeployArgs {
   export?: string
   migrationDir?: string
   noTypeId: boolean
-  send: boolean
 }
 
 function isMultisigFromInfo(fromInfo: FromInfo): fromInfo is MultisigScript {
@@ -90,11 +89,10 @@ subtask('contract:deploy')
     true,
   )
   .addParam('export', 'export transaction to file', '', paramTypes.path, true)
-  .addParam('send', 'send transaction directly', false, paramTypes.boolean, true)
   .addParam('no-type-id', 'not use type id deploy', false, paramTypes.boolean, true)
   .addParam('migration-dir', 'migration directory for saving json format migration files', '', paramTypes.path, true)
   .setAction(async (args: ContractDeployArgs, { config, run }) => {
-    const { name, from, feeRate, signer, binPath, noTypeId = false, send = false, migrationDir } = args
+    const { name, from, feeRate, signer, binPath, noTypeId = false, migrationDir } = args
     const { ckbChain } = config
 
     const fromInfos = parseFromInfoByCli(from)
@@ -190,9 +188,7 @@ subtask('contract:deploy')
       })
 
       writeFileSync(exportPath, JSON.stringify(exportData, null, 2))
-    }
-
-    if (send) {
+    } else {
       const txHash = await sendTx()
       console.info('deploy success, txHash: ', txHash)
       if (migrationDir) {
@@ -231,7 +227,6 @@ interface ContractUpgradeArgs {
   deployer?: string[]
   feeRate?: number
   export?: string
-  send: boolean
 }
 subtask('contract:upgrade')
   .addParam('name', 'name of the contract to upgrade', '', paramTypes.string, true)
@@ -262,9 +257,8 @@ subtask('contract:upgrade')
     true,
   )
   .addParam('export', 'export transaction to file', '', paramTypes.path, true)
-  .addParam('send', 'send transaction directly', false, paramTypes.boolean, true)
   .setAction(async (args: ContractUpgradeArgs, { config, run }) => {
-    const { name, feePayer, deployer, feeRate, signer, binPath, send = false, migrationDir } = args
+    const { name, feePayer, deployer, feeRate, signer, binPath, migrationDir } = args
     const { ckbChain } = config
 
     const deployerInfos = deployer ? parseFromInfoByCli(deployer) : []
@@ -388,9 +382,7 @@ subtask('contract:upgrade')
       })
 
       writeFileSync(exportPath, JSON.stringify(exportData, null, 2))
-    }
-
-    if (send) {
+    } else {
       const txHash = await sendTx()
       console.info('deploy success, txHash: ', txHash)
 
--- a/packages/core/src/builtin-tasks/contract.ts
+++ b/packages/core/src/builtin-tasks/contract.ts
@@ -1,6 +1,6 @@
 import { execSync } from 'node:child_process'
 import read from 'read'
-import { existsSync, writeFileSync, cpSync, rmSync } from 'node:fs'
+import { existsSync, writeFileSync, cpSync, rmSync, mkdirSync } from 'node:fs'
 import path from 'node:path'
 import { KuaiError } from '@ckb-js/kuai-common'
 import { ERRORS } from '../errors-list'
@@ -16,6 +16,7 @@ import { task, subtask } from '../config/config-env'
 import { paramTypes } from '../params'
 import { getUserConfigPath } from '../project-structure'
 import { getGenesisScriptsConfig } from '../util/chain'
+import { generateMigrationFileName, findMigrationByDir } from '../util/contract-migration'
 
 task('contract').setAction(async () => {
   execSync('kuai contract --help', { stdio: 'inherit' })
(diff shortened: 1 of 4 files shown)


packages/core/src/builtin-tasks/contract.ts
```diff
@@ -16,6 +16,18 @@ import { task, subtask } from '../config/config-env'
 import { paramTypes } from '../params'
 import { getUserConfigPath } from '../project-structure'
 import { getGenesisScriptsConfig } from '../util/chain'
+import { Hash } from '@ckb-lumos/base'
+
+type ContractMigration = {
+  cell_recipes: Array<{
```
Keith-CY, line 22: What does `recipe` mean in `migration`

packages/core/src/builtin-tasks/contract.ts
```diff
@@ -16,6 +16,18 @@ import { task, subtask } from '../config/config-env'
 import { paramTypes } from '../params'
 import { getUserConfigPath } from '../project-structure'
 import { getGenesisScriptsConfig } from '../util/chain'
+import { Hash } from '@ckb-lumos/base'
```
Keith-CY, line 19:
```suggestion
import type { Hash } from '@ckb-lumos/base'
```

packages/core/src/builtin-tasks/contract.ts
```diff
@@ -198,6 +210,181 @@ subtask('contract:deploy')
     return tx
   })
 
+interface ContractUpgradeArgs {
+  name?: string
+  binPath?: string
+  migrationPath: string
+  signer?: 'ckb-cli' | 'ckb-cli-multisig'
+  feepayer?: string[]
```
Keith-CY, line 218: `feepayer` could be `feePayer`

packages/core/src/builtin-tasks/contract.ts
```diff
@@ -198,6 +210,181 @@ subtask('contract:deploy')
...
+  deployer?: string[]
+  feeRate?: number
+  export?: string
+  send: boolean
+}
+subtask('contract:upgrade')
+  .addParam('name', 'name of the contract to be deployed', '', paramTypes.string, true)
```
Keith-CY, line 225:
```suggestion
  .addParam('name', 'name of the contract to upgrade', '', paramTypes.string, true)
```

packages/core/src/builtin-tasks/contract.ts
```diff
@@ -198,6 +210,181 @@ subtask('contract:deploy')
...
+  .addParam('export', 'export transaction to file', '', paramTypes.path, true)
+  .addParam('send', 'send transaction directly', false, paramTypes.boolean, true)
+  .setAction(async (args: ContractUpgradeArgs, { config, run }) => {
+    const { name, feepayer, deployer, feeRate, signer, binPath, send = false, migrationPath } = args
+    const { ckbChain } = config
+
+    const feepayerInfos = feepayer ? parseFromInfoByCli(feepayer) : []
+    const deployerInfos = deployer ? parseFromInfoByCli(deployer) : []
+
+    const migrationFile: ContractMigration = await (async () => {
```
Keith-CY, line 253: Which method makes it `async/await`

packages/core/src/builtin-tasks/contract.ts
```diff
@@ -198,6 +210,181 @@ subtask('contract:deploy')
...
+                ),
+              ),
+              require_first_n: fromInfo.R,
+              threshold: fromInfo.M,
+            },
+          })
+        }
+      })
+
+      writeFileSync(exportPath, JSON.stringify(exportData, null, 2))
```
Keith-CY, line 377: In what case will it be exported and sent simultaneously?

packages/core/src/contract.ts
```diff
@@ -83,4 +83,59 @@ export class ContractDeployer {
...
+    send: () => Promise<Hash>
+  }> {
+    const { feeRate = 1000 } = options || {}
+    const contractBin = readFileSync(contractBinPath)
+
+    const typeIdScript = await (async () => {
+      const tx = await this.#rpc.getTransaction(contract.txHash)
+
+      return tx.transaction.outputs[contract.index].type!
+    })()
```
Keith-CY, line 113: Why not run these methods directly but wrap them into an IIFE?

packages/core/src/builtin-tasks/contract.ts
```diff
@@ -198,6 +210,181 @@ subtask('contract:deploy')
...
+
+    const conrtactBinPath = await (async () => {
+      // check bin path is absolute path
+      if (binPath && path.isAbsolute(binPath)) {
+        return binPath
+      }
+
+      if (binPath) {
+        return path.join(process.cwd(), binPath)
+      }
```
Daryl-L, line 293: Isn't it better to coding like follow?

```typescript
      // check bin path is absolute path
      if (binPath) {
        if (binPath && path.isAbsolute(binPath)) {
          return binPath
        }

        return path.join(process.cwd(), binPath)
      }
```

packages/core/src/builtin-tasks/contract.ts
```diff
@@ -198,6 +210,181 @@ subtask('contract:deploy')
...
+      deployerInfos[0],
+      feepayerInfos[0],
+      {
+        txHash: migration.tx_hash,
+        index: migration.index,
+        dataHash: migration.data_hash,
+      },
+      { feeRate },
+    )
+    const { tx } = deployTx
```
Daryl-L, line 342:
```suggestion
    const { tx } = await contractDeployer.upgrade(
      conrtactBinPath,
      deployerInfos[0],
      feepayerInfos[0],
      {
        txHash: migration.tx_hash,
        index: migration.index,
        dataHash: migration.data_hash,
      },
      { feeRate },
    )
```

packages/core/src/builtin-tasks/contract.ts
```diff
@@ -198,6 +210,181 @@ subtask('contract:deploy')
...
+  feeRate?: number
+  export?: string
+  send: boolean
+}
+subtask('contract:upgrade')
+  .addParam('name', 'name of the contract to be deployed', '', paramTypes.string, true)
+  .addParam('bin-path', 'path of contract bin file', '', paramTypes.path, true)
+  .addParam('migration-path', 'path of migration file')
+  .addParam('feepayer', 'address or multisig config of the transaction fee payer', '', paramTypes.string, false, true)
+  .addParam('deployer', 'address or multisig config of the contract deployer', '', paramTypes.string, false, true)
```
yanguoyu, line 229: I can't get the point of `feepayer` and `deployer`, could we directly use the `deployer` cell to provide the tx fee?

packages/core/src/builtin-tasks/contract.ts
PainterPuppets: This migrations file is generated by `capsule`, and I'm not quite sure what `cell_recipes` means

yanguoyu (approved):

Keith-CY:
Any update for the conversations?

Daryl-L (approved):

packages/core/src/builtin-tasks/contract.ts
PainterPuppets: Is this just an option for the user, or how about sending without exporting by default?

Keith-CY: IMO, either `send` or `export` is enough because I cannot imagine a case of `send & export`

Keith-CY: Any update on this conversation

PainterPuppets: ok I'll update it to send by default in the absence of export

Keith-CY (approved):

Issue #325, opened 2023-06-18
Inproper wrapper of response


Keith-CY:
![image](https://github.com/ckb-js/kuai/assets/7271329/acebc951-06ec-444d-8139-39be8f2724ff)

```
#/load/{address}/addresses
{
  "code": "200",
  "data": {
    "code": "200",
    "data": [
      {
        "key": "ckb",
        "value": "ckt1qrejnmlar3r452tcg57gvq8patctcgy8acync0hxfnyka35ywafvkqgpgc5378xav7zf3e93ewcf40rcaskwfdf5qq9wz2f4"
      }
    ]
  }
}
```

Keith-CY:
How is it going

PR #378, opened 2023-07-19
fix: mvp dapp inproper wrapper of response

PainterPuppets:
fix: https://github.com/ckb-js/kuai/issues/325


Daryl-L (approved):

yanguoyu (approved):

packages/samples/mvp-dapp/src/app.controller.ts
```diff
@@ -60,7 +60,7 @@ router.get<never, { path: string; address: string }>('/load/:address/:path', asy
   )
   const value = recordModel.load(`data.${params.path}`)
   if (value) {
-    ctx.ok(MvpResponse.ok(MvpResponse.ok(value)))
+    ctx.ok(MvpResponse.ok(value))
```
Keith-CY, line 63: So the nested response is caused by business code, and the response is not wrapped by `kuai` automatically? 🤔

Maybe we can schedule the automatic wrapping of responses at the most outside as 

```mermaid
flowchart TD
	S[Service]
	C[Controller]
	I[Interceptor]
  W[Wrapped response]
  S -- res --> C -- res --> I --> W
  S -- res --> I --> W
```

Keith-CY (approved):

Issue #400, opened 2023-07-27
`env-paths` is required in `@ckb-js/kuai-common` but not specified in package.json


Keith-CY:
`env-paths` is used at https://github.com/ckb-js/kuai/blob/develop/packages/common/src/constant.ts#L1, but it's not listed in https://github.com/ckb-js/kuai/blob/develop/packages/common/package.json

It's weird that `env-paths` is not specified in any package.json, so it's not a phantom dependency. Why isn't it blamed on building locally 🤔

Keith-CY:
It's a phantom dependency introduced by `node-gyp` at https://github.com/ckb-js/kuai/blob/8f21bcaf354464ef637fc1d04dbdcab6ba872223/package-lock.json#L9929

Keith-CY:
Fixed by https://github.com/ckb-js/kuai/pull/401

PR #401, opened 2023-07-28
fix(common): add dependency env-paths to common package

Daryl-L:
#400

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5528,7 +5528,6 @@
       "version": "2.2.1",
       "resolved": "https://registry.npmjs.org/env-paths/-/env-paths-2.2.1.tgz",
       "integrity": "sha512-+h1lkLKhZMTYjog1VEpJNG7NZJWcuc2DDk/qsqSTRRCOXiLjeQ1d1/udrUGhqMxUgAlwKNZ0cf2uqan5GLuS2A==",
-      "dev": true,
       "engines": {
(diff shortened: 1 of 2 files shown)


yanguoyu (approved):

Keith-CY (approved):

homura:
Downgrading to 2.2.1 may resolve the question, this is because the [env-paths](https://github.com/sindresorhus/env-paths/releases/tag/v3.0.0) was changed to pure ESM in v3, it is incompatible with the Kuai's [commonjs](https://github.com/ckb-js/kuai/blob/develop/tsconfig.json#L28) module system

to import it, we can use the `import` API

```ts
const envPaths = await import('env-paths').then(res => res.default)
envPaths('...')
```

Daryl-L:
> Downgrading to 2.2.1 may resolve the question, this is because the [env-paths](https://github.com/sindresorhus/env-paths/releases/tag/v3.0.0) was changed to pure ESM in v3, it is incompatible with the Kuai's [commonjs](https://github.com/ckb-js/kuai/blob/develop/tsconfig.json#L28) module system
> 
> 
> 
> to import it, we can use the `import` API
> 
> 
> 
> ```ts
> 
> const envPaths = await import('env-paths').then(res => res.default)
> 
> envPaths('...')
> 
> ```
> 
> 

Let me have a try

PR #402, opened 2023-07-28
chore: Add option for redis

yanguoyu:
Add the options parameter for connecting Redis.

### DIFF
--- a/packages/samples/mvp-dapp/kuai.config.js
+++ b/packages/samples/mvp-dapp/kuai.config.js
@@ -6,12 +6,22 @@
 
 require('dotenv').config()
 
+let redisOpt = undefined
+if (process.env.REDIS_OPT) {
+  try {
+    redisOpt = JSON.parse(process.env.REDIS_OPT)
+  } catch (error) {
+    //ignore error, if error redisOpt will be undefined
+  }
+}
+
 module.exports = {
   host: process.env.HOST,
   port: process.env.PORT,
   network: 'testnet',
   redisPort: process.env.REDIS_PORT,
   redisHost: process.env.REDIS_HOST,
+  redisOpt,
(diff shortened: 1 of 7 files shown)


packages/samples/mvp-dapp/src/type-extends.d.ts
```diff
@@ -14,5 +14,6 @@ declare module '@ckb-js/kuai-core' {
     lumosConfig?: Config | 'aggron4' | 'lina'
     redisPort?: number
     redisHost?: string
+    redisOpt?: object
```
Keith-CY, line 17: The type can be more specific and imported from https://github.com/redis/ioredis/blob/main/lib/redis/RedisOptions.ts#L184

packages/samples/mvp-dapp/kuai.config.js
```diff
@@ -6,12 +6,20 @@
 
 require('dotenv').config()
 
+let redisOpt = undefined
+if (process.env.REDIS_OPT) {
+  try {
+    redisOpt = JSON.parse(process.env.REDIS_OPT)
+  } catch (error) {}
```
Keith-CY, line 13: Better to add a comment to indicate the error is ignored by design

Keith-CY (approved):

packages/samples/mvp-dapp/kuai.config.js
Keith-CY: Ah it should be the same as https://github.com/ckb-js/kuai/pull/402/files#diff-7592d8066442ab2d0309fc0c5dd84caf99a8aa00a1e78e446ae0e47b9c11ad1aR3-R11

Daryl-L (approved):

PR #416, opened 2023-08-04
fix: list missing dependencies in package.json

Keith-CY:
This PR appends missing dependencies in packages.json of each sub package. Most missing dependencies are phantom dependencies that are installed by other external sub-packages.


Keith-CY:
@Daryl-L @yanguoyu @PainterPuppets please have a review

packages/common/package.json
```diff
@@ -8,8 +8,11 @@
     "test": "jest"
   },
   "dependencies": {
-    "@ckb-lumos/lumos": "0.20.0",
     "env-paths": "2.2.1",
     "undici": "5.22.1"
+  },
+  "devDependencies": {
+    "@ckb-lumos/base": "0.20.0",
+    "@ckb-lumos/lumos": "0.20.0"
```
Daryl-L, line 16: Why to move these two dependencies to `devDependencies`?

Keith-CY: They are only referenced in https://github.com/ckb-js/kuai/blob/develop/packages/common/src/util.ts#L1-L2 as type definitions and will be wiped out after compilation. In other words, they are not used for production.

packages/io/package.json
```diff
@@ -43,8 +43,11 @@
     "rxjs": "7.8.1"
   },
   "devDependencies": {
+    "@ckb-lumos/base": "0.20.0",
+    "@ckb-lumos/rpc": "0.20.0",
     "@types/koa-compose": "3.2.5",
     "@types/koa-router": "7.4.4",
-    "koa": "2.14.2"
+    "koa": "2.14.2",
+    "reflect-metadata": "0.1.13"
```
yanguoyu, line 51: we have added `@ckb-lumos/lumos` at 37 line, these devDependencies are unnecessary.

packages/models/package.json
```diff
@@ -32,12 +32,15 @@
     "@ckb-js/kuai-common": "0.0.1-alpha.0",
     "@ckb-js/kuai-io": "0.0.1-alpha.0",
     "@ckb-lumos/base": "0.20.0",
+    "@ckb-lumos/bi": "0.20.0",
     "@ckb-lumos/codec": "0.20.0",
     "@ckb-lumos/experiment-tx-assembler": "0.20.0",
+    "@ckb-lumos/lumos": "0.20.0",
```
yanguoyu, line 38: we have added @ckb-lumos/lumos at the 38 line, other lumos dependencies may be unnecessary. Because lumos has include others package.

packages/io/package.json
Keith-CY: They will be phantom dependencies if they are not specified in the list. And the code will break once phantom dependencies are not installed along with `@ckb-lumos/lumos`

yanguoyu: `@ckb-lumos/lumos` has added all sub-packages of the Lumos to dependencies. When the sub-packages will not install?

Keith-CY: > `@ckb-lumos/lumos` has added all sub-packages of the Lumos to dependencies. When the sub-packages will not install?

Future is hard to predicate 😑 here're 3 cases for the explanation.

1. A sub-package is removed from `@ckb-lumos/lumos`: It will fail to import the sub-package because it's removed from `@/ckb-lumos/lumos` and won't be installed implicitly
2. Incompatibility introduced by implicitly installed sub-package. Using an implicitly installed package is similar to using an unexported method of a package. It's an internal implementation and is not intended for public usage. The compatibility won't be guaranteed by the maintainer. Say a method named `A` is imported from a `phantom package@v0.0.1`, but it's removed since `phantom package@v0.0.2`, the code breaks because `phantom package` is not explicitly specified in the package.json.
3. If `phantom dependency` is acceptable, will you skip installing `lodash` if `lumos` uses it?

yanguoyu (approved):

packages/common/package.json
Keith-CY: > I think it may not only be referenced for type definition in the future.

They can be listed in `dependencies` when used for production.
Listing unused packages in `dependencies` because they may be used in the future is somehow over-optimized.

Keith-CY: I found this conversation has been marked solved, does that mean PR has been reviewed?

Daryl-L (approved):

packages/common/package.json
Daryl-L: Yes, and I reviewed other dependencies, and no other missing found.

PR #421, opened 2023-08-07
feat(docker-node): update version of undici manually

Daryl-L:
1. Remove the undici from dependencies, and use fetch which is a built-in module in node 18.
2. The node version in some sub packages should be larger than 18

#415

### DIFF
--- a/packages/models/src/contract/contract.ts
+++ b/packages/models/src/contract/contract.ts
@@ -1,4 +1,3 @@
-import { request } from 'undici'
 import { ActorURI, CallResponse, MessagePayload } from '../actor'
 import { ParamsMissException as MethodMissException } from '../exceptions'
 import { Store } from '../store'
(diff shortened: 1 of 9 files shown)


Keith-CY:
`fetch` has been natively supported in Node.js LTS.

Ref:
  - fetch: https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API#browser_compatibility

yanguoyu (approved):

packages/models/src/contract/contract.ts
```diff
@@ -36,7 +35,7 @@ export class Contract<
         status: Status.ok,
         message: {
           pattern: payload.pattern,
-          value: res.body.json(),
+          value: res.json(),
```
Keith-CY: `res.json()` returns a promise that should be resolved before being returned.

Ref:
- json: https://developer.mozilla.org/en-US/docs/Web/API/Response/json

Keith-CY (approved):

PR #423, opened 2023-08-08
fix: use railway env name

yanguoyu:
Cause of using `REDIS_` env name will not run success with the `railway`. So I change the env name to the `railway`'s default env name.

### DIFF
--- a/packages/samples/mvp-dapp/kuai.config.js
+++ b/packages/samples/mvp-dapp/kuai.config.js
@@ -15,13 +15,27 @@ if (process.env.REDIS_OPT) {
   }
 }
 
+// fallback to REDISUSER due to https://github.com/ckb-js/kuai/pull/423#issuecomment-1668983983
+const REDIS_USER = redisOpt?.username ?? process.env.REDUSUSER
+const REDIS_PASSWORD = redisOpt?.password ?? process.env.REDISPASSWORD
+const REDIS_HOST = process.env.REDIS_HOST ?? process.env.REDISHOST
+const REDIS_PORT = process.env.REDIS_PORT ?? process.env.REDISPORT
+
+const redisAuth = REDIS_USER && REDIS_PASSWORD ? { username: REDIS_USER, password: REDIS_PASSWORD } : undefined
+
 module.exports = {
   host: process.env.HOST,
   port: process.env.PORT,
   network: 'testnet',
-  redisPort: process.env.REDIS_PORT,
-  redisHost: process.env.REDIS_HOST,
-  redisOpt,
+  redisPort: REDIS_PORT,
+  redisHost: REDIS_HOST,
+  redisOpt:
+    redisOpt || redisAuth
+      ? {
+          ...redisOpt,
+          ...redisAuth,
+        }
+      : undefined,
   jest: {
     preset: 'ts-jest',
     testEnvironment: 'node',


Keith-CY:
> Cause of using `REDIS_` env name will not run success with the `railway`. So I change the env name to the `railway`'s default env name.

Have you found the reason why `REDIS_` doesn't work?

yanguoyu:
> > Cause of using `REDIS_` env name will not run success with the `railway`. So I change the env name to the `railway`'s default env name.
> 
> Have you found the reason why `REDIS_` doesn't work?

From discord: 
> Right. One service has an env var with connection string, which references env vars (user, pass etc) from another service
http://${{Clickhouse.CLICKHOUSE_USER}}:${{Clickhouse.CLICKHOUSE_PASSWORD}}@hostname/${{Clickhouse.CLICKHOUSE_DB}}
becomes
http://:@hostname/
Yeah, I can't get it to work in any other way except changing the env var value to something else, then back to the template I use
After the change, it properly replaces the tokens
Tried restarting the service, same issue.
Yeah, it seems to happen no matter the template/setup
And honestly, makes sense. It's the clasic cat and mouse issue, where you don't know which variables got to be computed first

I don't know when they will fix this. Changing the env name is a faster way to build a successful railway template.

Keith-CY:
> > > Cause of using `REDIS_` env name will not run success with the `railway`. So I change the env name to the `railway`'s default env name.
> > 
> > 
> > Have you found the reason why `REDIS_` doesn't work?
> 
> From discord:
> 
> > Right. One service has an env var with connection string, which references env vars (user, pass etc) from another service
> > http://${{Clickhouse.CLICKHOUSE_USER}}:${{Clickhouse.CLICKHOUSE_PASSWORD}}@hostname/${{Clickhouse.CLICKHOUSE_DB}}
> > becomes
> > http://:@hostname/
> > Yeah, I can't get it to work in any other way except changing the env var value to something else, then back to the template I use
> > After the change, it properly replaces the tokens
> > Tried restarting the service, same issue.
> > Yeah, it seems to happen no matter the template/setup
> > And honestly, makes sense. It's the clasic cat and mouse issue, where you don't know which variables got to be computed first
> 
> I don't know when they will fix this. Changing the env name is a faster way to build a successful railway template.

Got the point, although I still don't know why `copy and paste the options again` solves it. But glad to see a workaround for it.

However, this PR changes too many fields in the project from the following considerations:

1. only the MVP dapp is affected by this issue, so the update is expected to be limited in the MVP dapp
2. it can be fixed in the framework level by patching `environments` as `export const REDIS_PORT = +(process.env.REDIS_PORT ?? process.env.REDISPORT ?? DEFAULT_REDIS_PORT)` with a comment, instead of updating all `REDIS_x` into `REDISx` in a weird spell
3. Unnecessary replacement at https://github.com/ckb-js/kuai/pull/423/files#diff-18da939443858e7c61d4f9a6e8b559f11cc19105af98f0eaa117d86a196ecc5eR13-R16

yanguoyu:
Updated with only sample-mvp.

packages/samples/mvp-dapp/kuai.config.js
```diff
@@ -7,9 +7,18 @@
 require('dotenv').config()
```
Keith-CY: What if using

```typescript
// fallback to REDISUSER due to https://github.com/ckb-js/kuai/pull/423#issuecomment-1668983983
const REDIS_USER = process.env.REDIS_USER ?? process.env.REDUSUSER
const REDIS_PASSWORD = process.env.REDIS_PASSWORD ?? process.env.REDISPASSWORD
const REDIS_OPT = process.env.REDIS_OPT ?? process.env.REDISOPT
const REDIS_HOST = process.env.REDIS_HOST ?? process.env.REDISHOST
const REDIS_PORT = process.env.REDIS_PORT ?? process.env.REDISPORT
```

`REDIS_*` are kept because they are defined in the `env.example`(https://github.com/ckb-js/kuai/blob/develop/packages/samples/mvp-dapp/.env.example) and work locally. `REDIS*`s only work for railway

yanguoyu: `REDIS_OPT` includes `username` and `password`. And `railway` does not include the `REDISOPT ` variable.
And we are not defining `REDIS_USER` and `REDIS_PASSWORD`.

Do you mean to write them together and add an annotation?

Keith-CY: > Do you mean to write them together and add an annotation?

Almost yes, I would like to keep the names of variables consistent so it's clear that they are paired.

Otherwise, I would like to figure out why 
`port` is `REDIS_PORT ?? REDISPORT` but 
`username` is `REDISUSER`, not `REDIS_USER`, and why 
`redis option` is `REDIS_OPT` but not `REDISOPT`

Keith-CY: Another issue is that

```
REDIS_OPT = {"username":"${{Redis.REDISUSER}}","password":"${{Redis.REDISPASSWORD}}"}
```
is defined in railway configuration, but it is overridden by

```
  redisOpt = {
    username: process.env.REDISUSER,
    password: process.env.REDISPASSWORD,
  }
  redisOpt = {
    ...JSON.parse(process.env.REDIS_OPT),
    ...redisOpt,
  }
```

yanguoyu: If they define `REDISUSER` and `REDISPASSWORD`. It means they want to use `REDISUSER` and `REDISPASSWORD`.

Keith-CY: > If they define `REDISUSER` and `REDISPASSWORD`. It means they want to use `REDISUSER` and `REDISPASSWORD`.

If it's a reply to https://github.com/ckb-js/kuai/pull/423#discussion_r1286760953

> Another issue is that
> 
> ```
> REDIS_OPT = {"username":"${{Redis.REDISUSER}}","password":"${{Redis.REDISPASSWORD}}"}
> ```
> is defined in railway configuration, but it is overridden by
> 
> ```
>   redisOpt = {
>     username: process.env.REDISUSER,
>     password: process.env.REDISPASSWORD,
>   }
>   redisOpt = {
>     ...JSON.parse(process.env.REDIS_OPT),
>     ...redisOpt,
>   }
> ```

My point is that `REDIS_OPT` can be left empty by default in railway because `username` and `password` of `REDIS_OPT` are overridden by railway's default `REDISUSER` and `REDISPASSWORD`

yanguoyu: > My point is that REDIS_OPT can be left empty by default in railway because username and password of REDIS_OPT are overridden by railway's default REDISUSER and REDISPASSWORD

Of cause, it is used to resolve the bug that `REDIS_OPT ` is not correct. But if the user deploys without `railway`, they can define `REDIS_OPT` without `REDISUSER ` and `REDISPASSWORD `.

Keith-CY (approved):

Daryl-L (approved):

Issue #334, opened 2023-06-21
Support caching and loading deployment information of contracts


Keith-CY:
Now there's a convention in the MVP DApp tutorial, as mentioned at https://github.com/ckb-js/kuai/issues/306

> The deployment information should be generated automatically as a facility for the backend. It will be implemented soon, now we have to set it manually at `profile/contract/deployed/contracts.json`.

> The deployment information of the online demo can be found at [ebffba1/packages/samples/mvp-dapp/contract/deployed_demo](https://github.com/ckb-js/kuai/tree/ebffba1bc73255fa3ccf6d85300129c70bf88f47/packages/samples/mvp-dapp/contract/deployed_demo)

With this convention, the deployed contracts can be referenced by `getDeployedContracts`(https://github.com/ckb-js/kuai/blob/develop/packages/samples/mvp-dapp/src/utils/index.ts#L5-L23) for `Pattern` and `Transaction View`.

It would be great to inject the deployment information of contracts into runtime/pipeline automatically to reduce the mental burden of referencing contracts.

For this purpose, I would suggest that the deployment information should be generated by `kuai contract command` in a similar format as @Daryl-L designed at https://github.com/ckb-js/kuai/issues/235#issuecomment-1519093262 at a specific path, and an internal utility function should be exposed to loading deployment information for application runtime, and pipeline of the task system(https://github.com/ckb-js/kuai/issues/13). The utility function could be designed in the `contract model`(https://github.com/ckb-js/kuai/blob/develop/packages/models/src/contract/contract.ts)

Is this feature feasible and helpful to developers? @Daryl-L @PainterPuppets @yanguoyu

yanguoyu:
It would be better if we can output the contract information when using `kuai contract deploy`.
IMO, we can expose the loading contract utils for developers, but it's unnecessary to load at runtime, it can be decided by developers.

Keith-CY:
> It would be better if we can output the contract information when using `kuai contract deploy`. IMO, we can expose the loading contract utils for developers, but it's unnecessary to load at runtime, it can be decided by developers.

Outputing deployment information along with `kuai contract deploy` is a good idea.

But what's the drawback of loading contracts automatically?

PainterPuppets:
We currently have a migration file, so maybe we could use the migration file for this configuration?

Migration file look like that
```json
{
  "cell_recipes": [
    {
      "name": "sample-dapp",
[...]

PR #425, opened 2023-08-09
feature(contract): add the contract manager to manage the deployment config

Daryl-L:
1. add contract loader to load contract deployment config from file
2. add contract manager to manage deployment config which is loaded
3. use the contract manager in bin node

#334

### DIFF
--- a/packages/common/src/contract/interface.ts
+++ b/packages/common/src/contract/interface.ts
@@ -1,0 +1,19 @@
+import type { CellDep, Script } from '@ckb-lumos/lumos'
+
+export type ContractDeploymentInfo = {
+  name: string
+  path: string
+  cellDeps?: { name?: string; cellDep?: CellDep }[]
+  scriptBase: Omit<Script, 'args'>
+} & CellDep
+
+export interface ContractLoader {
+  load: () => ContractDeploymentInfo[]
+  write: (info: ContractDeploymentInfo[]) => void
+}
+
+export interface ContractManager {
+  contracts: ContractDeploymentInfo[]
+  updateContract: (info: ContractDeploymentInfo, loader?: ContractLoader) => void
+  write: () => void
+}
--- a/packages/common/src/contract/interface.ts
+++ b/packages/common/src/contract/interface.ts
@@ -11,9 +11,3 @@ export interface ContractLoader {
   load: () => ContractDeploymentInfo[]
   write: (info: ContractDeploymentInfo[]) => void
 }
-
-export interface ContractManager {
-  contracts: ContractDeploymentInfo[]
-  updateContract: (info: ContractDeploymentInfo, loader?: ContractLoader) => void
-  write: () => void
-}
--- a/packages/common/src/contract/interface.ts
+++ b/packages/common/src/contract/interface.ts
@@ -1,8 +1,9 @@
 import type { CellDep, Script } from '@ckb-lumos/lumos'
+import { Path } from './path'
 
 export type ContractDeploymentInfo = {
   name: string
-  path: string
+  path: Path
   cellDeps?: ({ name: string } | { cellDep: CellDep })[]
   scriptBase: Omit<Script, 'args'>
 } & CellDep
--- a/packages/common/src/contract/interface.ts
+++ b/packages/common/src/contract/interface.ts
@@ -3,7 +3,7 @@ import type { CellDep, Script } from '@ckb-lumos/lumos'
 export type ContractDeploymentInfo = {
   name: string
   path: string
-  cellDeps?: { name?: string; cellDep?: CellDep }[]
+  cellDeps?: ({ name: string } | { cellDep: CellDep })[]
   scriptBase: Omit<Script, 'args'>
 } & CellDep
 
--- a/packages/common/src/contract/kuai.loader.ts
+++ b/packages/common/src/contract/kuai.loader.ts
@@ -1,6 +1,6 @@
 import type { ContractDeploymentInfo, ContractLoader } from './interface'
 import fs from 'node:fs'
-import { validateName, validatePath, validateScriptBase, validatecellDep } from './validator'
+import { validateCellDeps, validateName, validatePath, validateScriptBase, validateCellDep } from './validator'
 
 export class KuaiContractLoader implements ContractLoader {
   constructor(private path: string) {}
@@ -12,8 +12,12 @@ export class KuaiContractLoader implements ContractLoader {
     if (!Array.isArray(info)) {
       throw new Error(`info must be an array`)
     }
-    for (const key in info) {
-      validatecellDep(validatePath(validateName(validateScriptBase(validatecellDep(info[key])))))
+    for (const item of info) {
+      validateCellDeps(item)
+      validatePath(item)
+      validateName(item)
+      validateScriptBase(item)
+      validateCellDep(item)
     }
 
     fs.writeFileSync(this.path, Buffer.from(JSON.stringify(info)))
--- a/packages/common/src/contract/kuai.loader.ts
+++ b/packages/common/src/contract/kuai.loader.ts
@@ -1,5 +1,6 @@
 import type { ContractDeploymentInfo, ContractLoader } from './interface'
 import fs from 'node:fs'
+import { validateName, validatePath, validateScriptBase, validatecellDep } from './validator'
 
 export class KuaiContractLoader implements ContractLoader {
   constructor(private path: string) {}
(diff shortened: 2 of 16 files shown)


Keith-CY:
How is it going

Daryl-L:
> How is it going

under testing, will be ok tonight

packages/common/src/contract/kuai.loader.ts
```diff
@@ -0,0 +1,11 @@
+import { ContractDeploymentInfo, ContractLoader } from './interface'
+import fs from 'fs'
+
+export class KuaiContractLoader implements ContractLoader {
+  constructor(private path: string) {}
+
+  load = (): ContractDeploymentInfo[] =>
+    fs.existsSync(this.path) ? JSON.parse(fs.readFileSync(this.path, 'utf-8')) ?? [] : []
```
yanguoyu, line 9: Should we add try to catch unexpected exceptions?

packages/common/src/contract/manager.ts
```diff
@@ -0,0 +1,44 @@
...
+    })
+  }
+
+  getContract = (name: string): ContractDeploymentInfo | undefined => {
+    const contract = this.#contracts.get(name)
+
+    if (contract) {
+      const [info] = contract
+      return info
+    }
```
yanguoyu: How about 
```
return contract?.[0]
```

packages/common/src/contract/kuai.loader.ts
```diff
@@ -0,0 +1,11 @@
+import { ContractDeploymentInfo, ContractLoader } from './interface'
+import fs from 'fs'
```
Keith-CY:
```suggestion
import type { ContractDeploymentInfo, ContractLoader } from './interface'
import fs from 'node:fs'
```

packages/common/src/contract/kuai.loader.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export class KuaiContractLoader implements ContractLoader {
+  constructor(private path: string) {}
+
+  load = (): ContractDeploymentInfo[] =>
+    fs.existsSync(this.path) ? JSON.parse(fs.readFileSync(this.path, 'utf-8')) ?? [] : []
+
+  write = (info: ContractDeploymentInfo[]) => fs.writeFileSync(this.path, Buffer.from(JSON.stringify(info)))
```
Keith-CY: Validation should be added if content is totally overriden.

packages/common/src/contract/manager.ts
```diff
@@ -0,0 +1,44 @@
+import { ContractDeploymentInfo, ContractLoader } from './interface'
```
Keith-CY:
```suggestion
import type { ContractDeploymentInfo, ContractLoader } from './interface'
```

packages/common/src/contract/manager.ts
```diff
@@ -0,0 +1,44 @@
...
+      loader.load().forEach((contract) => {
+        this.#contracts.set(contract.name, [contract, loader])
+      })
+    })
+  }
+
+  updateContract = (info: ContractDeploymentInfo, loader?: ContractLoader) => {
+    const contract = this.#contracts.get(info.name)
+    if (contract) {
+      const [_, loader] = contract
```
Keith-CY: This `loader` shadows `loader` passed in.

packages/common/src/contract/kuai.loader.ts
Daryl-L: I think, developers who use this function should catch this exception because of wrong content in file provided by themselves.

packages/common/src/contract/kuai.loader.ts
Daryl-L: > Validation should be added if content is totally overriden.

After thinking about it, I have no idea what should be validated in this?

Keith-CY: > > Validation should be added if content is totally overriden.
> 
> After thinking about it, I have no idea what should be validated in this?

1. iterate each contract object in the json
2. check if each mandatory field exists, e.g. name
3. check if value of each field is valid: e.g. cellDeps should be an array of `Record<'name', string>` or `CellDep`

Daryl-L: > > > Validation should be added if content is totally overriden.
> > 
> > 
> > After thinking about it, I have no idea what should be validated in this?
> 
> 1. iterate each contract object in the json
> 2. check if each mandatory field exists, e.g. name
> 3. check if value of each field is valid: e.g. cellDeps should be an array of `Record<'name', string>` or `CellDep`

OK, I see. What I think about was wrong before.

yanguoyu (approved):

packages/common/src/contract/interface.ts
```diff
@@ -0,0 +1,13 @@
+import type { CellDep, Script } from '@ckb-lumos/lumos'
+
+export type ContractDeploymentInfo = {
+  name: string
+  path: string
```
PainterPuppets: The path here should be a relative path? after all, every personal computer on the absolute path are not quite the same, if it is a relative path, then which path should be used as root?

Daryl-L: Absolute path maybe suitable here I think.

Keith-CY: > Absolute path maybe suitable here I think.

It's better to append arguments along with the conclusion

Keith-CY: Status update: Absolute and relative paths will be supported simultaneously

packages/common/src/contract/kuai.loader.ts
```diff
@@ -1,11 +1,21 @@
...
   load = (): ContractDeploymentInfo[] =>
     fs.existsSync(this.path) ? JSON.parse(fs.readFileSync(this.path, 'utf-8')) ?? [] : []
 
-  write = (info: ContractDeploymentInfo[]) => fs.writeFileSync(this.path, Buffer.from(JSON.stringify(info)))
+  write = (info: ContractDeploymentInfo[]) => {
+    if (!Array.isArray(info)) {
+      throw new Error(`info must be an array`)
+    }
+    for (const item of info) {
+      validatecellDep(validatePath(validateName(validateScriptBase(validatecellDep(item)))))
```
Keith-CY: This usage is a bit weird. Each checker returns `item` directly, which is called monad, so it can be refactored into

```js
item
|> validatecellDep
|> validatePath
|> validateName
```

But the pipeline operator in JavaScript has not been approved, so it is usually in OOP form

```js
item.validatecellDep().validatePath().validateName()
```

However, in this case the `item` will not be updated in each checker. It can be simply executed separately

```js
validatecellDep(item)
validatePath(item)
validateName(item)
```

---

BTW, `validatecellDep` is called twice in this list. Maybe one of them should be `validateCellDeps`

packages/common/src/contract/validator.ts
```diff
@@ -0,0 +1,86 @@
...
+          }
+        }
+      })
+    ) {
+      throw new Error(`cellDeps must be an array of { name: string } or { cellDep: CellDep }`)
+    }
+  }
+
+  return info
+}
```
Keith-CY: Nested too much.

1. `!Array.isArray(info.cellDeps)` can be at the top
2. the condition of `if` at line 10 is complex and should be a function

packages/common/src/contract/validator.ts
```diff
@@ -0,0 +1,86 @@
...
+
+export const validatePath = (info: ContractDeploymentInfo) => {
+  if (!info.path) {
+    throw new Error(`path is required`)
+  }
+
+  return info
+}
+
+export const validatecellDep = (info: ContractDeploymentInfo) => {
```
Keith-CY: `validatecellDep` should be `validateCellDep`

packages/common/src/contract/validator.ts
Daryl-L: > Nested too much.
> 
> 1. `!Array.isArray(info.cellDeps)` can be at the top
> 2. the condition of `if` at line 10 is complex and should be a function

This validator only validates the `cellDeps` only when the contract has dependent cells. So if the `info.cellDeps` is undefined, the function will return directly.

Keith-CY (approved):

packages/common/src/contract/path.ts
```diff
@@ -0,0 +1,17 @@
+import path from 'node:path'
+
+export class Path {
+  constructor(private _path: string) {}
+
+  get path(): string {
+    if (path.isAbsolute(this._path)) {
+      return this._path
+    } else {
+      return path.resolve(process.cwd(), this._path)
```
Keith-CY: It requires users to run the command in the root directory. What if using something like https://github.com/ckb-js/kuai/blob/8e3cdb59fca7eddabeb592d8d169872d62e21919/packages/core/src/util/packageInfo.ts#L19

packages/common/src/util.ts
```diff
@@ -6,6 +6,7 @@ import path from 'node:path'
 import fs from 'node:fs'
 import { PATH } from './constant'
 import { pipeline } from 'node:stream/promises'
+import findup from 'find-up'
```
Keith-CY, line 9: `find-up` should be added in `dependencies`

PR #445, opened 2023-08-22
feature/support write contract info

PainterPuppets:
reslove: https://github.com/ckb-js/kuai/issues/334

### DIFF
--- a/packages/core/src/builtin-tasks/contract.ts
+++ b/packages/core/src/builtin-tasks/contract.ts
@@ -2,7 +2,7 @@ import { execSync } from 'node:child_process'
 import read from 'read'
 import { existsSync, writeFileSync, cpSync, rmSync, mkdirSync } from 'node:fs'
 import path from 'node:path'
-import { KuaiError } from '@ckb-js/kuai-common'
+import { KuaiError, KuaiContractLoader, ContractManager, configPath } from '@ckb-js/kuai-common'
 import { ERRORS } from '../errors-list'
 import type { FromInfo, MultisigScript } from '@ckb-lumos/common-scripts'
 import { parseFromInfo } from '@ckb-lumos/common-scripts'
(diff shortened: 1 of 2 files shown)


yanguoyu (approved):

packages/core/src/builtin-tasks/contract.ts
```diff
@@ -213,6 +217,26 @@ subtask('contract:deploy')
...
+          hashType: hashType,
+        },
+        outPoint: {
+          txHash: txHash,
+          index: '0x' + index.toString(16),
+        },
+        depType: depType,
+      })
+
+      contractManager.write()
```
Daryl-L, line 239: Use `await` maybe better I think.

Keith-CY:
Please add a brief description or link this PR to the requirement issue in the PR message, so we can have a peek with `git log` command

packages/core/src/contract.ts
```diff
@@ -40,6 +40,8 @@ export class ContractDeployer {
   ): Promise<{
     tx: Transaction
     index: number
+    hashType: 'type' | 'data1'
```
Keith-CY: Why not add `data` in the type

packages/core/src/contract.ts
```diff
@@ -104,6 +108,8 @@ export class ContractDeployer {
     tx: Transaction
     index: number
     dataHash: string
+    hashType: 'type' | 'data1'
```
Keith-CY: Why not add `data` in the type

Keith-CY:
Tests should be added

packages/core/src/builtin-tasks/contract.ts
PainterPuppets: The "write" function doesn't seem to be async function, will `await` work?

Daryl-L: > The "write" function doesn't seem to be async function, will `await` work?
It's my fault, I had the wrong understanding before.

Daryl-L (approved):

PR #457, opened 2023-08-30
feat: Init sudt backend by kuai init

yanguoyu:


Daryl-L:
seems that test failed?

Keith-CY (approved):

Keith-CY:
I think we can approve and merge this PR quickly because it's simply generated by the `cli` as a codebase.

Daryl-L (approved):

Issue #462, opened 2023-09-07
Alias `pattern` as `filter`


Keith-CY:
This issue comes from feedback from a user

---

The `Pattern` decorator works exactly as `Filter` does in a functional context, so aliasing it as `filter` would help users to understand its function intuitively.

Keith-CY:
PR: https://github.com/ckb-js/kuai/pull/466

PR #466, opened 2023-09-08
feat: alias Pattern to Filter for better understanding in functional

Keith-CY:
context

Ref: https://github.com/ckb-js/kuai/issues/462


packages/models/src/utils/decorator/provider.ts
```diff
@@ -144,3 +144,14 @@ export const DefaultLock =
 export const Omnilock = (): ClassDecorator => DefaultLock('OMNILOCK')
 
 export const Secp256k1Lock = (): ClassDecorator => DefaultLock('SECP256K1_BLAKE160')
+
+/*
+ * Alias pattern to filter for better understanding in functional context
+ */
+export {
+  Pattern as Filter,
```
yanguoyu, line 152: Why not change the `Pattern` name when defined?

Keith-CY: > Why not change the `Pattern` name when defined?

It will be a breaking change that some users who have tried it will be bothered.

Keith-CY: `Pattern` can be considered a term used in an internal context, in fact, it was referenced in several documents and discussions. `Filter` can be considered as an user-oriented keyword for developers who are familiar to functional programming

yanguoyu (approved):

Daryl-L (approved):

PR #488, opened 2023-09-19
feat: Add omnilock and sudt actor.

yanguoyu:


packages/io/src/cor.ts
```diff
@@ -23,7 +29,7 @@ export class CoR<ContextT extends object = Record<string, never>> implements ICo
         err: rej,
       } as Context<ContextT>
 
-      compose([this._exceptionHandler, ...this._middlewares])(ctx)
+      compose([this._exceptionHandler, ...this._middlewares, notFoundMiddleware])(ctx)
```
PainterPuppets, line 32: The `CoR` is just a chain of responsibility and doesn't feel like it should be related to `http`

However, currently there is no case to use `CoR` except for `http`,  if there are other case in the future and then replace `http error` to the `adapter` is okay!

packages/io/src/controller.ts
```diff
@@ -132,7 +131,6 @@ export class BaseController {
 
         return route.middleware(ctx, next)
       } else {
-        ctx.err(new NotFound())
```
Keith-CY, line 135: Why remove the error message

packages/samples/sudt/src/actors/index.ts
```diff
@@ -4,3 +4,5 @@ const appRegistry = new Registry();
 appRegistry.load(__dirname);
 
 export { appRegistry };
+export { OmnilockModel } from './omnilock.model';
```
Keith-CY: Rule `semi: false` https://github.com/ckb-js/kuai/blob/develop/.prettierrc#L6 and https://github.com/ckb-js/kuai/blob/develop/packages/core/sample-projects/.prettierrc#L5 are inconsistent, I think it would be better to keep them the same.

packages/samples/sudt/src/actors/omnilock.model.ts
```diff
@@ -0,0 +1,110 @@
...
+    amount: HexString,
+  ): {
+    inputs: Cell[];
+    outputs: Cell[];
+    witnesses: string[];
+  } {
+    const CONFIG = getConfig();
+    const sudtCell: Cell = {
+      cellOutput: {
+        capacity: BI.from('14400000000').toHexString(),
```
Keith-CY: `14400000000` is a magic number

packages/samples/sudt/src/actors/sudt.model.ts
```diff
@@ -0,0 +1,149 @@
...
+      capacity: capacity.toString(),
+      sudtBalance: sudtBalance.toString(),
+    };
+  }
+
+  send(lockScript: Script, amount: HexString) {
+    const CONFIG = config.getConfig();
+    const sudtCell: Cell = {
+      cellOutput: {
+        capacity: BI.from('14400000000').toHexString(),
```
Keith-CY: `14400000000` is a magic number and why the capacity of recipient is set as a fixed value

packages/io/src/controller.ts
yanguoyu: When there are some controller routers, it will throw an error when the matching router is in the next controller. We'd better throw not found exception after all routers have matched.

packages/samples/sudt/src/response.ts
```diff
@@ -0,0 +1,24 @@
...
+  }
+
+  static ok<T>(data: T) {
+    return new SudtResponse('200', data);
+  }
+
+  static err<T>(data: T, code: string) {
+    return new SudtResponse(code, data);
+  }
+}
```
Daryl-L, line 24: I think maybe we can add it as a default response for developers to choose to the framework.

Keith-CY: How would it be used if it's a default response

Daryl-L: > How would it be used if it's a default response

I think we can define a default response type like this, and then developers could use it directly without designing another response class.

Keith-CY: > > How would it be used if it's a default response
> 
> I think we can define a default response type like this, and then developers could use it directly without designing another response class.

You may refer to a `BaseError` to extend, and the framework will handle `error` extended from `BaseError` automatically, e.g. output formatted error message, submit error to monitor, etc.

yanguoyu: I copied the `Response` from mvp-dapp. Shall we support the default response struct? If so, we'd better add a default middleware to transfer the response.

Keith-CY: > I copied the `Response` from mvp-dapp. Shall we support the default response struct? If so, we'd better add a default middleware to transfer the response.

It's a good idea but requires some work, maybe we should open an issue and add this feature in another PR

yanguoyu: > maybe we should open an issue and add this feature in another PR

I agree with that.

packages/samples/sudt/src/controllers/sudt.controller.ts
```diff
@@ -0,0 +1,42 @@
...
+import { SudtModel, appRegistry } from '../actors'
+import { Tx } from '../views/tx.view'
+import { getLock } from '../utils'
+import { SudtResponse } from '../response'
+
+@Controller('sudt')
+export default class SudtController extends BaseController {
+  @Post('/send')
+  async send(
+    @Body() { from, to, amount, typeArgs }: { from: string[]; to: string; amount: HexString; typeArgs: Hash },
```
Daryl-L, line 36: Why is the `from` array here?

packages/samples/sudt/src/response.ts
Keith-CY: Tracked by issue https://github.com/ckb-js/kuai/issues/491

packages/samples/sudt/src/controllers/sudt.controller.ts
Keith-CY: The account could be a set of addresses when connected to Neuron

Daryl-L: > The account could be a set of addresses when connected to Neuron

I see.

Daryl-L: > The account could be a set of addresses when connected to Neuron

So the `from` is an array only when the Neuron is connected, right?

Keith-CY: > > The account could be a set of addresses when connected to Neuron
> 
> So the `from` is an array only when the Neuron is connected, right?

All wallets connected will be considered sets of addresses, even if there is only 1 address bound to the account, e.g. metamask.

So, if the dapp is connected to metamask, the `from` will be `[address_of_metamask]`

packages/samples/sudt/src/controllers/omnilock.controller.ts
```diff
@@ -0,0 +1,37 @@
...
+  @Post('/mint')
+  async mint(@Body() { from, to, amount }: { from: string; to: string; amount: HexString }) {
+    if (!from || !to || !amount) {
+      throw new BadRequest('undefined body field: from, to or amount')
+    }
+
+    const omniLockModel = appRegistry.findOrBind<OmnilockModel>(
+      new ActorReference('omnilock', `/${getLock(from).args}/`),
+    )
+    const result = omniLockModel.mint(getLock(to), amount)
```
Daryl-L, line 34: What does the lock `to` mean here?

packages/samples/sudt/src/controllers/sudt.controller.ts
Keith-CY: It can be consider an abstract of account, all accounts connected to the dapp should be sets of addresses

packages/samples/sudt/src/controllers/omnilock.controller.ts
Keith-CY: > What does the lock `to` mean here?

The recipient to get minted token

Keith-CY (approved):

packages/samples/sudt/src/actors/sudt.model.ts
```diff
@@ -0,0 +1,152 @@
...
+import { InternalServerError } from 'http-errors'
+import { BI, utils, config } from '@ckb-lumos/lumos'
+import { MIN_SUDT_WITH_OMINILOCK, TX_FEE } from '../const'
+
+/**
+ * add business logic in an actor
+ */
+@ActorProvider({ ref: { name: 'sudt', path: `/:lockArgs/:typeArgs/` } })
+@TypeFilter()
+@Sudt()
```
Daryl-L, line 30: From the design of the `Model`, we need to use `Lock` to decorate the `SudtModel`. There must be a `lockscript` in the instance. As we only use `omnilock`, we should add the `@Omnilock()` here.

packages/samples/sudt/src/actors/sudt.model.ts
```diff
@@ -0,0 +1,152 @@
...
+export class SudtModel extends JSONStore<Record<string, never>> {
+  constructor(
+    @Param('args') args: string,
```
Daryl-L, line 33:
```suggestion
    @Param('lockArgs') lockArgs: string,
    @Param('typeArgs') typeArgs: string,
```

packages/samples/sudt/src/actors/sudt.model.ts
```diff
@@ -0,0 +1,152 @@
...
+    _schemaOption?: void,
+    params?: {
+      states?: Record<OutPointString, never>
+      chainData?: Record<OutPointString, UpdateStorageValue>
+      cellPattern?: CellPattern
+      schemaPattern?: SchemaPattern
+    },
+  ) {
+    super(undefined, { ...params, ref: ActorReference.newWithFilter(SudtModel, `/${args}/`) })
```
Daryl-L, line 42:
```suggestion
    super(undefined, { ...params, ref: ActorReference.newWithFilter(SudtModel, `/${lockArgs}/${typeArgs}/`) })
```

packages/samples/sudt/src/actors/sudt.model.ts
yanguoyu: This model will collect all the sudt cells without a lock script filter. We need to collect some locks owned by one user.

Daryl-L: > This model will collect all the sudt cells without a lock script filter. We need to collect some locks owned by one user.

The `Model` is designed to related `lock`, here I mark it down. https://github.com/ckb-js/kuai/issues/504#issuecomment-1756666591
Or we should change the design to support it.

yanguoyu: After merge https://github.com/ckb-js/kuai/pull/509. Sudt model can listen on type script without lock script.

yanguoyu: I have change ```@ActorProvider({ ref: { name: 'sudt', path: `/:lockArgs/:typeArgs/` } }) ``` to ```@ActorProvider({ ref: { name: 'sudt', path: `/:args/` } })```

packages/io/src/cor.ts
yanguoyu: Okay, here will replace `http error` with `adapt` when there are other cases.

packages/samples/sudt/src/actors/sudt.model.ts
yanguoyu: This has been fixed by https://github.com/ckb-js/kuai/pull/509

packages/samples/sudt/src/actors/sudt.model.ts
yanguoyu: This has been fixed by https://github.com/ckb-js/kuai/pull/509

packages/samples/sudt/src/actors/sudt.model.ts
yanguoyu: This has been fixed by https://github.com/ckb-js/kuai/pull/509

Keith-CY:
Please have a review @Daryl-L @PainterPuppets

Daryl-L (approved):

PR #509, opened 2023-10-12
fix: Resource binding can bind one of lock or type, or bind lock and type

### LINKED ISSUE #504: Some problem in developing the sUDT manager
1. The consideration for transaction information was overlooked. In the development of the sUDT management platform, a requirement arose to manage the transaction history, necessitating the synchronization of on-chain transactions. However, within Kuai, there is no built-in functionality for syncing transactions. Consequently, it becomes necessary to independently implement the synchronization of transaction records in the script and store them in the database.

2. The development experience with Kuai and Lumos in tandem is quite problematic. Take, for example, the generation of an sUDT, where Lumos provides the function `issueToken`. However, this function lacks the capability to pass cell information as the issuer. This renders the cell information synchronized by Kuai somewhat redundant. Regardless of whether Kuai provides cell information in the Model, Lumos needs to re-query through script or other account data, such as the `payFee` function. To enhance the Kuai development experience, it is essential for Lumos, as a framework for Nervos, to make certain adaptations.

yanguoyu:
This PR is used to fix https://github.com/ckb-js/kuai/issues/504#issuecomment-1756929777.

### DIFF
--- a/packages/io/src/chain-source/nervos.ts
+++ b/packages/io/src/chain-source/nervos.ts
@@ -23,7 +23,11 @@ export class NervosChainSource implements ChainSource {
     return header.number
   }
 
-  getAllLiveCellsWithWitness = async (lockScript?: Script, typeScript?: Script, startBlock?: BI, endBlock?: BI) => {
+  getAllLiveCellsWithWitness = async (
+    scripts: Partial<Record<'lock' | 'type', Script>>,
+    startBlock?: BI,
+    endBlock?: BI,
+  ) => {
     const res: (CKBComponents.IndexerCell & { witness: string })[] = []
 
     startBlock = startBlock ?? BI.from(0)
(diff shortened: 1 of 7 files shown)


packages/io/src/chain-source/nervos.ts
```diff
@@ -20,25 +23,43 @@ export class NervosChainSource implements ChainSource {
     return header.number
   }
 
-  getAllLiveCellsWithWitness = async (
-    lockScript: Script,
-    typeScript?: Script | undefined,
-    startBlock?: BI,
-    endBlock?: BI,
-  ) => {
+  getAllLiveCellsWithWitness = async (lockScript?: Script, typeScript?: Script, startBlock?: BI, endBlock?: BI) => {
```
Keith-CY: Two objects with named fields would be clearer, otherwise `undefined` should be passed in explicitly.

```ts
getAllLiveCellsWithWitnesses = async (scripts: Partial<Record<'lock' | 'type', Script>>, blockRange?: Partial<Record<'start' | 'end', BI>> => {}
```

packages/io/src/chain-source/nervos.ts
```diff
@@ -20,25 +23,43 @@ export class NervosChainSource implements ChainSource {
...
+        withData: true,
+      }
+    } else if (typeScript) {
+      scriptFilterOption = {
+        script: typeScript,
+        scriptType: 'type',
+        withData: true,
+      }
+    }
+    if (!scriptFilterOption) throw new Error('One of lock script and type script is required at least')
```
Keith-CY:
```suggestion
    if (!scriptFilterOption) throw new Error('One of lock script or type script is required at least')
```

Keith-CY (approved):

Keith-CY:
Please have a review @Daryl-L @PainterPuppets

Daryl-L (approved):

PR #511, opened 2023-10-12
fix: jest is missing in dev dependencies of template

Keith-CY:
jest is specified because jest.config.ts is listed in the template


yanguoyu (approved):

Keith-CY:
@PainterPuppets @Daryl-L please have a review

Daryl-L (approved):

PR #512, opened 2023-10-12
Feature/sudt backend

Daryl-L:
backend API of sudt manager.


Keith-CY:
Online preview is available at https://sudt-management-ui-production.up.railway.app/ with Neuron at https://github.com/nervosnetwork/neuron/pull/2880#issuecomment-1835643188

